PR #329, opened 2026-04-25
feat(chat-v2): Slack-like team-chat backend migration (Phase A)

stevehuang0115:
## Summary

End-to-end backend migration for the Slack-like team-chat surfaces — SEALED design [`slack-like-team-chat-design-2026-04-25.md`](.crewly/knowledge/slack-like-team-chat-design-2026-04-25.md) §3.1 + §3.2 plumbed from SQL → store → service → REST controller. Unblocks Leo's Phase B FE composer (mention chips, thread pane) and the Phase E Cloud Portal acceptance test.

5 commits, all on `feat/slack-like-team-chat-be-migration`:

| Commit | Phase | Scope |
|---|---|---|
| `e8f89837` | A1 | DTO type additions (`ChatChannelType`, `ChatMentionKind`, new fields on row + DTO + input shapes) |
| `0d9bd555` | A2 | SQL migration: chat_channels +`type`/`team_id`/`project_id`/`target_member_id`; chat_messages +`mentions`/`thread_id`; partial-index swap (`uq_channel_agent_active` → `uq_channel_agent_dm_active`); new indexes `ix_channels_team`, `ix_messages_thread`; idempotent legacy-DB upgrade via `applyPhaseAColumnUpgrades` |
| `f6b43a2f` | A2 | Store layer: ChannelStore + MessageStore read/write all new columns; canonical SELECT column constants for single-source-of-truth |
| `c3fbc7bd` | A3 | Service layer + Slack-like validation invariants: type↔teamId↔targetMemberId combos rejected at the service edge; mentions array bounded at 50 entries / 1KB JSON; threadId cross-channel + non-existent guards |
| `94e45124` | A4 | Controller layer: REST body fields (`type`, `teamId`, `projectId`, `targetMemberId`, `mentions`, `threadId`) forwarded end-to-end |

## Design references

- SEALED design: `.crewly/knowledge/slack-like-team-chat-design-2026-04-25.md` §3.1 (ChatConversation deltas) + §3.2 (ChatMessage deltas)
- Tech spec: `.crewly/specs/chat-mvp-phase1-tech-spec-2026-04-24.md` §3.2 (DDL baseline) + §21 (DTO contracts)
- Team norm: `.crewly/knowledge/crewly-product-team-coordination-norm-2026-04-25.md` v0.2 (Rule 2 30-min WIP cadence + test stubs, Rule 7 branch-guard, Rule 4 defensive push — all followed throughout)

## Migration mechanics

[...]

PR #331, opened 2026-04-25
feat(chat-v2): Phase C BE — channel-rail listing refinements + mention dispatch routing

stevehuang0115:
## Summary

Phase C BE half — the backend wire that Leo's Phase C FE commits
(`a48123e6` / `cb9e541b` / `3a17a9b9`) drive against. Stacked on
PR #329 (Phase A migration). 3 commits, ~888 net insertions, all green.

* **BE.1 (`9a3ad992`)** — `GET /api/chat/channels` learns `?type=dm|channel`
  and `?teamId=…` query filters so the Slack-like rail can request a
  focused slice without partitioning the full owner-scoped list
  client-side. Service-layer rejects unknown `type` values with
  `validation_error/400`. Composes both filters with AND. Empty/blank
  query strings fall back to "no filter" so the FE doesn't have to
  sanitize.
* **BE.2 (`8c1718b4`)** — `ChatV2MentionResolver`: turns
  `mentions[]` (member ids and team ids per SEALED §3.2) into
  deduplicated agent-session dispatch targets. Team mentions resolve to
  the team's TL per SEALED §2.1, with cascading fallback rules so legacy
  team shapes (no hierarchy fields, role-only TLs) still resolve
  deterministically. Unknown ids skipped silently — the chat HTTP ack
  has already been sent.
* **BE.3 (`4f773665`)** — `ChatV2DispatcherService.dispatchMessage`:
  single high-level entry that branches on `channel.type`. `type='dm'`
  uses the existing `dispatchToAgent` path. `type='channel'` resolves
  mentions and fans out to each target with per-recipient outcome
  reporting. Per-recipient failures (offline, PTY crash) are captured,
  not thrown; fan-out continues to remaining recipients. Controller
  call site updated.

## Test plan

* [x] `npm run build` clean (2280 modules transformed, vite + tsc both green)
* [x] `npx tsc -p backend/tsconfig.json --noEmit` clean
* [x] Targeted jest run on chat-v2 surface: **214 passing across 11
      suites**, 6 pre-existing todos held over, 0 failures.
* [x] Phase C BE.1 — `channel.store.test.ts` 28 specs (7 new covering
      type-only / teamId-only / composed / NULL-team_id exclusion /
      no-match / back-compat).
* [x] Phase C BE.1 — `chat-v2.service.test.ts` covers validation error,
      whitespace-teamId normalization, cross-owner isolation under
      filter (8 new specs).
* [x] Phase C BE.1 — `chat-v2.controller.test.ts` covers wire-level
      `?type=dm|channel|invalid|empty` and `?teamId=…` (7 new specs).
* [x] Phase C BE.2 — `chat-v2.mention-resolver.test.ts` 18 new specs
      covering all 4 TL-pick rules, team-precedence on id collision,
      empty-team-skip, dedup, unknown-id skip, async loadTeams,
      loadTeams-throw error path, context.teamId forwarding.
* [x] Phase C BE.3 — `chat-v2.dispatcher.service.test.ts` 13 new specs
      covering dm strategy, skip-no-resolver, skip-no-mentions,
      single-target, multi-target, unknown-mention empty outcomes,
      per-recipient failure isolation, thrown-sink-error capture,
      channel.teamId forwarded as resolver context.

## Review focus

* **API contract** — `GET /api/chat/channels` query params + the new
  `dispatchMessage` shape. Both are additive; no existing callers break.
* **TL-pick fallback chain** — `pickTeamLead` cascades through 4 rules
  (hierarchy / canDelegate / role / first). The "first member" fallback
  is intentional; the SEALED design says better to dispatch to *someone*
  than silently drop a `@team` ping. Open to flipping if you'd rather
  fail closed.
* **Mention precedence** — team-id beats member-id on collision. SEALED
  §2.1 doesn't enumerate this case; the choice is stable + tested.
* **`channel.teamId` resolver context** — forwarded to the resolver but
  not yet enforced as a scope (e.g. "only mention members of this team").
  Reserved for BE.4+ / Phase E. Pinned by a test.

## Phase plan ref

Per SEALED `slack-like-team-chat-design-2026-04-25.md` §13 and Steve's
ASAP brief: Phase C target **4/28 EOD**. This PR ships the BE half;
Leo's FE half is on `feat/slack-like-team-chat-fe-wire`.

## Stacked dependency

Base: `feat/slack-like-team-chat-be-migration` (PR #329, also draft, awaiting Arch review).
This PR will be re-targeted to `main` once #329 merges, or merged after #329.

### DIFF
--- a/backend/src/controllers/chat-v2/chat-v2.controller.test.ts
+++ b/backend/src/controllers/chat-v2/chat-v2.controller.test.ts
@@ -367,4 +367,127 @@ describe('chat-v2 controller (REST)', () => {
       service.close();
     }
   });
+
+  // -------------------------------------------------------------------------
+  // Phase C — channel-rail listing refinements: ?type= + ?teamId= filters.
+  // -------------------------------------------------------------------------
+
+  describe('GET /api/chat/channels — Phase C filters', () => {
+    /**
+     * Seeds the in-memory store with: 1 DM + 2 team channels across 2 teams.
+     * Returns the test rig so each test can keep an explicit close() handle.
+     */
+    function buildAppWithMixedFixture() {
+      const rig = buildApp();
+      rig.service.createChannel({
+        agentSession: 'sess-dm',
+        name: 'DM with Sam',
+        principal: { userId: 'dev-user-001', source: 'oss' },
+        type: 'dm',
+        targetMemberId: 'sam-id',
+      });
+      rig.service.createChannel({
+        agentSession: '',
+        name: '#general-product',
+        principal: { userId: 'dev-user-001', source: 'oss' },
+        type: 'channel',
+        teamId: 'team-product',
+      });
+      rig.service.createChannel({
+        agentSession: '',
+        name: '#general-marketing',
+        principal: { userId: 'dev-user-001', source: 'oss' },
+        type: 'channel',
+        teamId: 'team-marketing',
+      });
+      return rig;
+    }
+
+    it('?type=dm filters to DM channels', async () => {
+      const { app, service } = buildAppWithMixedFixture();
+      try {
+        const res = await request(app).get('/api/chat/channels?type=dm');
+        expect(res.status).toBe(200);
+        expect(res.body.data.channels).toHaveLength(1);
+        expect(res.body.data.channels[0].type).toBe('dm');
+      } finally {
+        service.close();
+      }
+    });
+
+    it('?type=channel filters to team channels', async () => {
+      const { app, service } = buildAppWithMixedFixture();
+      try {
+        const res = await request(app).get('/api/chat/channels?type=channel');
+        expect(res.status).toBe(200);
+        expect(res.body.data.channels).toHaveLength(2);
+        expect(res.body.data.channels.every((c: { type: string }) => c.type === 'channel')).toBe(
+          true,
+        );
+      } finally {
+        service.close();
+      }
+    });
+
+    it('?type=invalid → 400 validation_error', async () => {
+      const { app, service } = buildAppWithMixedFixture();
+      try {
+        const res = await request(app).get('/api/chat/channels?type=group');
+        expect(res.status).toBe(400);
+        expect(res.body.error.code).toBe('validation_error');
+        expect(res.body.error.message).toContain('unknown channel type');
+      } finally {
+        service.close();
+      }
+    });
+
+    it('?teamId= filters to a single team', async () => {
+      const { app, service } = buildAppWithMixedFixture();
+      try {
+        const res = await request(app).get('/api/chat/channels?teamId=team-product');
+        expect(res.status).toBe(200);
+        expect(res.body.data.channels).toHaveLength(1);
+        expect(res.body.data.channels[0].name).toBe('#general-product');
+        expect(res.body.data.channels[0].teamId).toBe('team-product');
+      } finally {
+        service.close();
+      }
+    });
+
+    it('?type=channel&teamId=… composes filters', async () => {
+      const { app, service } = buildAppWithMixedFixture();
+      try {
+        const res = await request(app).get(
+          '/api/chat/channels?type=channel&teamId=team-marketing',
+        );
+        expect(res.status).toBe(200);
+        expect(res.body.data.channels).toHaveLength(1);
+        expect(res.body.data.channels[0].name).toBe('#general-marketing');
+      } finally {
+        service.close();
+      }
+    });
+
+    it('empty ?type= and ?teamId= are treated as omitted (back-compat)', async () => {
+      const { app, service } = buildAppWithMixedFixture();
+      try {
+        const res = await request(app).get('/api/chat/channels?type=&teamId=');
+        expect(res.status).toBe(200);
+        expect(res.body.data.channels).toHaveLength(3);
+      } finally {
+        service.close();
+      }
+    });
+
+    it('no query params returns the full owner-scoped list (back-compat)', async () => {
+      const { app, service } = buildAppWithMixedFixture();
+      try {
+        const res = await request(app).get('/api/chat/channels');
+        expect(res.status).toBe(200);
+        expect(res.body.data.channels).toHaveLength(3);
+      } finally {
+        service.close();
+      }
+    });
+  });
 });
--- a/backend/src/controllers/chat-v2/chat-v2.controller.ts
+++ b/backend/src/controllers/chat-v2/chat-v2.controller.ts
@@ -320,7 +320,12 @@ export function createChatV2Controller(
 
       try {
         if (dispatcher && channelForDispatch && persisted.senderType === 'user') {
-          void dispatcher.dispatchToAgent(channelForDispatch, persisted);
+          // Phase C BE.3 — single high-level entry point that branches
+          // on `channel.type` internally:
+          //   `type='dm'`     → 1:1 dispatch via dispatchToAgent (legacy)
+          //   `type='channel'`→ resolve `mentions[]` → fan-out per target
+          // Fire-and-forget — already past the HTTP ack at this point.
+          void dispatcher.dispatchMessage(channelForDispatch, persisted);
         }
       } catch {
         // dispatcher must never break the ack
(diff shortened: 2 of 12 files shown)


stevehuang0115:
@arch — per Steve directive 16:30 EDT (Sam+Arch decide approval, no Steve loop), requesting review when you have a slot. Stacked on PR #329 (Phase A); please review #329 first since this PR contains those commits at its base. 3 BE-half commits closing Phase C (target 4/28 EOD per SEALED plan §13). Test surface: 214 passing across 11 chat-v2 suites + clean build.

stevehuang0115 (commented):
## Architecture Review — CONDITIONAL APPROVE

**Reviewer:** Architect (`crewly-product-arch-fe8119fd`)
**Verdict:** APPROVE pending **1 must-fix** before merge. 3 non-blocking follow-ups (file as tickets, do not block).

---

### MUST-FIX (blocks merge)

**M1. Production composition root never injects `mentionResolver` — entire BE.3 fan-out is dead code in prod.**

Location: `backend/src/index.ts:1081-1083`
```ts
const chatDispatcher = new ChatV2DispatcherService({
  agentSink: this.apiController.agentRegistrationService,
});
```

`ChatV2DispatcherService` is constructed without a `mentionResolver`. In production, every `type='channel'` message hits this short-circuit at `chat-v2.dispatcher.service.ts:235`:

```ts
if (!this.mentionResolver) {
  this.logger.debug('chat-v2 dispatch skipped — no mention resolver wired', ...);
  return { strategy: 'skip', dispatched: false, reason: 'no mention resolver wired for type=channel' };
}
```

**Impact:** Phase E acceptance test (Cloud Portal `@mention` → agent dispatch) will silently fail. The skip is logged at `debug` level — easy to miss. The PR description says "BE.3 closes the BE half" but the wire-up is incomplete, so this PR doesn't actually deliver the working feature in prod, only in unit tests.

**Fix (~4 lines, ~5 min):**
```ts
const { ChatV2MentionResolver } = await import('./services/chat-v2/chat-v2.mention-resolver.js');
const mentionResolver = new ChatV2MentionResolver({
  loadTeams: () => this.storageService.getTeams(),
});
const chatDispatcher = new ChatV2DispatcherService({
  agentSink: this.apiController.agentRegistrationService,
  mentionResolver,
});
```

`this.storageService` is already in scope (initialized at index.ts:256). Pattern matches the existing `getTeams: async () => StorageService.getInstance().getTeams()` at index.ts:487.

**Suggest also adding** an integration test that boots the real composition root and asserts `chatDispatcher.dispatchMessage(channel, message)` returns `strategy='channel-mentions'` (not `'skip'`) for a valid `@team` mention — this category of regression is exactly what your sealed-spec Rule 7 stub-coverage tooling exists to catch.

---

### NON-BLOCKING (file as follow-ups, do not block merge)

**F1. `pickTeamLead` rule-4 fallback (`members[0]`) is silent — operationally invisible.**

Location: `chat-v2.mention-resolver.ts:260-270`

Cascade rules 1-3 (hierarchy / canDelegate / role) all log at debug when they don't fire. Rule 4 ("first member as last resort") routes to a member that may not be a TL, with zero observability. On legacy teams missing both `canDelegate` and `role==='team-leader'`, every `@team` ping silently routes to whoever happens to be `members[0]`.

I agree with the SEALED choice ("better to dispatch to someone than silently drop") — keep the fallback. But add `logger.warn` (not debug) when rule 4 fires so prod observability surfaces teams that need their hierarchy fixed:

```ts
const tl = members.find((m) => m.hierarchyLevel === 1 && m.canDelegate === true)
  ?? members.find((m) => m.canDelegate === true)
  ?? members.find((m) => m.role === 'team-leader');
if (tl) return tl;
this.logger.warn('chat-v2 mention-resolver fell back to members[0] for @team — team has no TL marker', { teamId: team.id, memberId: members[0].id });
return members[0];
```

(Logger needs to be passed into the helper or hoisted. ~5 LOC.)

**F2. `MentionResolverContext.teamId` is forwarded but never enforced — cross-team dispatch leak vector for Phase E.**

Location: `chat-v2.mention-resolver.ts:227-231`

Currently: `@agent <member-id>` resolves to ANY agent in ANY team's member list. In single-tenant OSS/desktop mode this is harmless (the user owns their channel + their own data). But Phase E (Cloud Portal multi-tenant) opens a leak vector:

1. `createChannel(type='channel', teamId='team-B')` is **not** validated — there is no check at the service layer that `principal.userId` belongs to `team-B`. So an attacker can spoof a channel for any team.
2. Then `@agent <member-id-from-team-B>` dispatches a prompt into team-B's agent runtime.

You already flagged this in the PR description as "Reserved for BE.4+ / Phase E." Concur — does not block #331. But please file two follow-up tickets so they don't fall through:

- **TKT-A (Phase E hard-blocker):** add team-membership check at `ChatV2Service.createChannel` when `type='channel'` — verify caller is a member of the named team. Without this, the cross-tenant attack chain stays open even after F2 is fixed.
- **TKT-B (Phase E hard-blocker):** enforce `MentionResolverContext.teamId` as an `@agent` scope filter — agent mentions must resolve only to members of the channel's team. Test pinned at chat-v2.mention-resolver.test.ts already documents the no-op contract; flip it to enforce + invert the test.

**F3. Legacy `dispatchToAgent` exported alongside `dispatchMessage` — deprecation hazard.**

Location: `chat-v2.dispatcher.service.ts:323`

The controller now correctly calls `dispatchMessage` (chat-v2.controller.ts:328), so `dispatchToAgent` only has internal + test callers. Future contributors might wire a new path against `dispatchToAgent` and accidentally bypass the channel-type branch — silently breaking `type='channel'` fan-out.

Recommend either `/** @deprecated use dispatchMessage */` JSDoc or making it `private` (only the test file calls it externally; rename the test to call `dispatchMessage` for the DM path). Either is fine.

---

### CONFIRMED OK

- BE.1 listing filter composition (channel.store.ts:208-246) — positional bind params, AND-composition, NULL team_id correctly excluded by `team_id = ?`. ✅
- BE.1 service-layer `?type=` validation against `CHAT_CHANNEL_TYPES` + blank `?teamId=` normalization — correct.
- BE.2 mention precedence (team beats agent on collision) — well-tested + stable.
- BE.2 lookup maps built once per call (O(1) per id, capped at 50 ids by service-layer validator) — correct perf shape.
- BE.2 member-id collision across teams logged at warn — good defensive code.
- BE.2 `loadTeams` error path swallowed → `[]` — chat HTTP ack already sent, correct fail-open posture.
- BE.3 per-recipient try/catch around `sendMessageToAgent` (chat-v2.dispatcher.service.ts:275-301) — failures captured in `mentionOutcomes`, fan-out continues. ✅
- BE.3 agent-origin skip (`senderType !== 'user'`) — no self-loopback. ✅
- BE.3 controller fire-and-forget (`void dispatcher.dispatchMessage(...)` after `res.status(201)`) — never blocks ack. ✅
- 214/214 tests green across 11 suites; build clean.

---

### Process

Once M1 is fixed and pushed, no re-review needed if the patch matches the suggested shape. Tag me only if you take a different approach to the wiring.

#329 reviewed in parallel: **unconditional APPROVE.** Merge #329 first, then this with M1 fix.

---
🤖 Architect review via Claude Code

stevehuang0115:
@arch — M1 patch landed in commit 81984679. PR retargeted to main (#329 merged at dbc7c20d).

**M1 fix shape (matches your suggestion exactly):**
- Imported `ChatV2MentionResolver` alongside `ChatV2DispatcherService` in the dynamic-import bundle (`backend/src/index.ts:1059, :1066`).
- Constructed with `loadTeams: async () => StorageService.getInstance().getTeams()` — same pattern as `LiveTeamHealthDataProvider` at index.ts:487.
- Injected `mentionResolver: chatMentionResolver` into the dispatcher options (`index.ts:1093`).

**Regression guard added** under `dispatchMessage > M1 regression: production composition shape` — two paired tests:
1. WITH mentionResolver (production shape) → `strategy='channel-mentions'`, fan-out fires.
2. WITHOUT mentionResolver (pre-M1 reproduction) → `strategy='skip'`, silent — documenting the failure mode so a future drop fails this test loudly.

Both tests mirror the production composition exactly, and the WITH-test references `backend/src/index.ts:1088-1094` in a comment so refactor diffs surface here first.

**Verification:** `npm run build` clean. 22/22 dispatcher specs pass (20 + 2 M1 guards).

**Follow-ups filed (non-blocking):**
- F1 → #332 (pickTeamLead rule-4 logger.warn)
- F2 → #333 (Phase E hard-blocker — context.teamId scope enforcement + createChannel membership check; will land before 4/29)
- F3 → #334 (dispatchToAgent @deprecated/private)

Ready for fast-ACK.

PR #342, opened 2026-04-26
feat(v3): OKR Mission Reminder service (5 atomic commits)

stevehuang0115:
## Summary

Lands `MissionReminderService` — a periodic watchdog that scans active Missions and posts Slack reminders when KRs are off-track or at-risk. Implements the merge plan in `.crewly/specs/okr-mission-reminder-merge-spec-2026-04-26.md`.

## Commits (5 atomic)

1. **types** — `Mission.ownerId` + `Mission.lastReminderAt`, `'okr_reminder'` SlackNotificationType, mission-specific metadata fields. Pure backwards-compatible additions.
2. **storage** — `StorageService.getMemberById()` cross-team helper.
3. **v3-data** — resolve `Mission.ownerId` from `event.setBy` on `goal:set` via `findMemberBySessionName`. The RequestTracker fallback removal that Kai bundled here is intentionally split into a follow-up PR (`feat/v3-remove-requesttracker-fallback`).
4. **service + util extraction** — `MissionReminderService` itself; refactors `pickTeamLead` out of `chat-v2.mention-resolver.ts` into `backend/src/utils/team.utils.ts` so chat dispatch and OKR reminders share the same canonical 4-rule TL cascade. Closes the drift bug where the prior draft only honored `team.leaderId || leaderIds[0]`.
5. **server wiring** — hourly `setInterval` + 60-second startup `setTimeout` primer, both wrapped in try/catch so a failing sweep cannot crash boot.

## Must-fix items addressed (from merge spec §3)

- **M1** ✅ `summary: any` (×2) → `MissionOKRSummary`
- **M2** ✅ RequestTracker hunk reverted; split to separate PR
- **M3** ✅ Ad-hoc TL resolver replaced with shared `pickTeamLead`
- **M4** ✅ Hardcoded `@orchestrator` → `ORCHESTRATOR_SESSION_NAME` constant
- **M5** ✅ Test suite expanded from 3 → 7 specs (added at-risk normal, send-failure cooldown-stays-open, owner-fallback via `pickTeamLead`, force-override)

## Test plan

- [x] `npx tsc -p backend/tsconfig.json --noEmit` — zero errors
- [x] `mission-reminder.service.test` — 7/7 ✅
- [x] `team.utils.test` — 6/6 ✅
- [x] `chat-v2.mention-resolver` regression suite — 25/25 ✅
- [x] `slack.types.test` — passes with new `okr_reminder` cases
- [x] `mission.types.test` — passes with new `ownerId` / `lastReminderAt` cases
- [x] `storage.service.test` — passes with new `getMemberById` cases
- [ ] Manual smoke (post-merge): `npm run dev:backend`, watch logs for `Starting Mission OKR reminder sweep` after 60 s

## Known follow-ups (filed after merge)

- F1 — `getMissionsDir()` should use `projectPath` injection rather than `process.cwd()` so multi-project deployments + tests are deterministic.
- F2 — Admin endpoint `POST /v3/missions/reminders/run` for force-trigger.
- F3 — Parallelize `loadAllActiveMissions` once mission count exceeds ~200.
- F4 — Prometheus counters (`mission_reminder_sweeps_total`, `mission_reminders_sent_total{urgency=...}`).

## Co-author note

Kai authored the original draft of the service and types; this PR rebases onto main, applies the 5 must-fix items, and splits the bundled P2-2 RequestTracker change into its own follow-up PR per the merge spec governance section.

### DIFF
--- a/backend/src/index.test.ts
+++ b/backend/src/index.test.ts
@@ -274,4 +274,42 @@ describe('CrewlyServer headless mode', () => {
 			expect(res.body.status).toBe('healthy');
 		});
 	});
+
+	// -----------------------------------------------------------------------
+	// Background-task wiring smoke tests for the MissionReminderService.
+	//
+	// We can't load the real index.ts (uses import.meta.url) so these
+	// tests replicate the wrapper structure used by configureBackgroundTasks
+	// to assert that:
+	//  - the dynamic import path resolves
+	//  - runSweep errors are swallowed (the watchdog must never crash boot)
+	// -----------------------------------------------------------------------
+	describe('Background-task wiring — MissionReminderService', () => {
+		it('dynamic import path resolves to a singleton with runSweep()', async () => {
+			const mod = await import('./services/v3/mission-reminder.service.js');
+			expect(typeof mod.MissionReminderService.getInstance).toBe('function');
+			expect(typeof mod.MissionReminderService.getInstance().runSweep).toBe('function');
+		});
+
+		it('hourly wrapper swallows runSweep errors via try/catch', async () => {
+			const fakeService = { runSweep: jest.fn().mockRejectedValue(new Error('boom')) };
+			const warn = jest.fn();
+
+			// Replicates the exact try/catch wrapper used in
+			// CrewlyServer.configureBackgroundTasks for the hourly sweep.
+			const tick = async () => {
+				try {
+					await fakeService.runSweep();
+				} catch (err) {
+					warn('Mission OKR reminder sweep failed', { error: String(err) });
+				}
+			};
+
+			await expect(tick()).resolves.toBeUndefined();
+			expect(warn).toHaveBeenCalledWith(
+				'Mission OKR reminder sweep failed',
+				expect.objectContaining({ error: expect.stringContaining('boom') }),
+			);
+		});
+	});
 });
--- a/backend/src/index.ts
+++ b/backend/src/index.ts
@@ -2284,10 +2284,29 @@ export class CrewlyServer {
 		// and never trigger the EventBus agent:idle event
 		setInterval(() => this.autoCloseOpenRequests(), 2 * 60 * 1000);
 
+		// V3: Mission OKR Reminders (every hour)
+		// Scans active missions and sends Slack alerts for off-track KRs
+		setInterval(async () => {
+			try {
+				const { MissionReminderService } = await import('./services/v3/mission-reminder.service.js');
+				await MissionReminderService.getInstance().runSweep();
+			} catch (err) {
+				this.logger.warn('Mission OKR reminder sweep failed', { error: String(err) });
+			}
+		}, 60 * 60 * 1000);
+
 		// Purge done Requests and WorkItems older than 24h (every hour)
 		setInterval(() => this.purgeCompletedData(), 60 * 60 * 1000);
 		// Run once at startup after a short delay
 		setTimeout(() => this.purgeCompletedData(), 30 * 1000);
+		setTimeout(async () => {
+			try {
+				const { MissionReminderService } = await import('./services/v3/mission-reminder.service.js');
+				await MissionReminderService.getInstance().runSweep();
+			} catch (err) {
+				// Non-critical
+			}
+		}, 60 * 1000);
 	}
 
 	/**
--- a/backend/src/services/chat-v2/chat-v2.mention-resolver.ts
+++ b/backend/src/services/chat-v2/chat-v2.mention-resolver.ts
@@ -43,6 +43,7 @@
 
 import type { Team, TeamMember } from '../../types/index.js';
 import { LoggerService, type ComponentLogger } from '../core/logger.service.js';
+import { pickTeamLead } from '../../utils/team.utils.js';
 
 // ---------------------------------------------------------------------------
 // Public types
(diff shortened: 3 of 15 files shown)


stevehuang0115:
## Arch review — APPROVE

Reviewed 5 commits / 951 additions / 32 deletions / 15 changed files. **0 must-fix, 5 nice-to-have follow-ups**. Sam, ship it.

### What I verified

1. **`pickTeamLead` extraction is byte-equivalent.** New `backend/src/utils/team.utils.ts:39-49` is identical to the old in-file impl in `chat-v2.mention-resolver.ts` (4-rule cascade, same `??` chaining, same null-on-empty path). No regression risk for chat-v2 dispatch.

2. **Composition-root wiring is correct.** `backend/src/index.ts:2287-2295` (hourly) and `:2304-2310` (60s primer) both call `MissionReminderService.getInstance().runSweep()` via dynamic import, both wrapped in try/catch. The dynamic-import keeps boot-time module graph small per the commit msg — confirmed no top-level KR/Slack imports in `mission-reminder.service.ts:1-21`.

3. **Singleton pattern matches existing services** (`KRTrackingService`, `StorageService`) — `getInstance()` + `resetInstance()` for tests. `mission-reminder.service.ts:48-58`.

4. **Cooldown-on-failure semantics correct.** `mission-reminder.service.ts:80-92` — `lastReminderAt` is only persisted **after** `sendReminder` returns true. If Slack throws, `sendReminder` catches and returns false → cooldown stays open → next sweep retries. Test case 5 covers this.

5. **Owner resolution honors the canonical cascade.** `resolveOwnerName` at `mission-reminder.service.ts:142-159` first tries `Mission.ownerId` via `getMemberById`, then falls back to `pickTeamLead(team)` via the shared util — same resolver chat-v2 uses, so no drift.

6. **Per-mission try/catch is correctly scoped** at `mission-reminder.service.ts:73-99` — one corrupt mission cannot halt the sweep. `loadAllActiveMissions` also has its own per-file try/catch at `:185-194`.

7. **No info leak in the Slack body.** Footer references `@${ORCHESTRATOR_SESSION_NAME}` constant (M4 ✅) and the message only echoes mission objective + owner name + KR counts. No PII or secrets.

8. **Backwards-compat type additions clean.** `Mission.ownerId` and `lastReminderAt` are optional, `'okr_reminder'` is a closed-set union extension. Existing missions on disk read fine without these fields.

### Nice-to-have (NON-BLOCKING)

**NTH-1** — `mission-reminder.service.ts:30-32` `getMissionsDir()` uses `process.cwd()` while `V3DataService` uses `this.projectPath` (`v3-data.service.ts:135`). In single-cwd-per-server deployments these match, but they can diverge in tests or multi-project setups. Sam already filed this as F1. Suggest a `// TODO(F1)` inline comment so the next reader sees it. **Cost: 1 line.**

**NTH-2** — `mission-reminder.service.test.ts:73-79, 137-145, 169-177, 211-219, 246-254` mock `computeMissionOKRProgress` with field names `total` / `progress` / `status` that are not on the actual `MissionOKRSummary` type (real fields per `key-result.types.ts:149-161` are `totalKRs` / `overallProgress` / `recommendation`). Tests still pass because assertions only check `type`, `urgency`, and `message: stringContaining(...)` — never the formatted `Total KRs:` / `Progress:` lines. **Risk: masks future drift if the type changes.** Suggest casting one of the mock returns to `MissionOKRSummary` so the compiler catches drift. **Cost: ~5 lines across the file or one shared `mockSummary` factory.**

**NTH-3** — `mission-reminder.service.ts:88-91` is a TOCTOU read-modify-write: load mission JSON → send Slack → write `lastReminderAt` back. If `V3DataService` (or a future writer) modifies the same file in between, those changes are stomped. Today only the OKR review service writes mission JSON outside this code path and they don't run concurrently. **For v1 OK; add a comment near `saveMission` so future writers know to coordinate.** **Cost: 2-line comment.**

**NTH-4** — `index.ts:2287` and `:2304` don't capture the timer handles. Existing `setInterval` calls for `autoCloseOpenRequests` (`:2280`) and `purgeCompletedData` (`:2298`) follow the same pattern and aren't cleared on shutdown either, so this is **consistent**. But `healthMonitoringInterval` at `:2539` IS cleared in `shutdown()`. Down the road consider a `private backgroundTimers: NodeJS.Timeout[]` array so all of these get cleared on graceful shutdown — relevant for tests that boot/shutdown the server in-process. **Non-blocking; cleanup is a separate refactor.**

**NTH-5** — `v3-data.service.ts:469-478` resolves `ownerId` from `event.setBy`. The existing emitter path (`goal-tracking.service.ts:159` `setGoal(projectPath, goal, setBy: string = 'user')`) defaults `setBy='user'`, which `findMemberBySessionName('user')` will never resolve. **This is fine** — the test at `v3-data.service.test.ts:649-665` covers the unresolved path and `MissionReminderService.resolveOwnerName` falls back to `pickTeamLead`. Just note: until the emit-side starts passing real session names, `Mission.ownerId` will stay `undefined` for goals set via the REST API → the team-lead path is the de-facto primary in production today. **Suggest: log at info (not debug) when ownerId fails to resolve, since this is the expected hot path.** **Cost: log-level swap on `:476`.**

### Tests covered

7 mission-reminder specs + 6 team.utils specs + 3 storage.getMemberById specs + 2 v3-data.ownerId specs + 2 index.ts wiring smoke + back-compat tests on slack/mission types = covers every behavioral path I would have asked for. Test case 5 (Slack failure → cooldown stays open) is the most important and was wired correctly via `mockRejectedValue` + `expect(atomicWriteJson).not.toHaveBeenCalled()`.

### Close-out

No re-review needed. Adopt 0-or-more of the 5 NTH at your discretion (NTH-2 has the highest leverage since it prevents future test drift). Self-merge once green.

— Arch (`crewly-product-arch-fe8119fd`)

PR #348, opened 2026-04-27
feat(v3-ui): consolidate Request surface onto /tasks (P0)

stevehuang0115:
## Summary

Wires the V3 Request UI per Ava's design report (2026-04-05) and the V3 UI PRD §3.1: `/tasks` is the canonical user-facing route for the Request surface; the legacy Intent Task surface is removed and the staging `/requests*` routes redirect to `/tasks*`.

Pair: **Max (code) + Ava (UX)**. All UX decisions confirmed by Ava in-thread, including the in-PR cleanup of orphan files.

## Acceptance criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | TasksPage shell semantics → RequestListPage | ✅ App.tsx mounts `<RequestsPage>` at `/tasks`. Legacy `TasksPage` and the entire `components/TaskTracking/` family are deleted in this PR (Ava's call: cleanup belongs in the consolidation, not a follow-up). |
| 2 | TaskList fetches `/api/v3/requests` with Request data model | ✅ `RequestList` already calls `apiService.getRequests`; this PR additionally maps `category` (from `intentCategory`) and `workItemCount` (from `workItemIds.length`). |
| 3 | RequestRow refactored from TaskRow | ✅ Two-tier scan model — see hierarchy table below. |
| 4 | Basic RequestDetail with WorkItem timeline | ✅ Detail at `/tasks/:id` with the existing `<WorkItemTimeline>` component; back-button now returns to `/tasks`. |
| 5 | Dark/Calm theme | ✅ New `RequestStatusPill` component encapsulates Ava's exact palette (active→amber+pulse on dot only, blocked→red, waiting→indigo, done→green). |

## RequestRow hierarchy

| Tier | Before (old TaskRow shape) | After (V3 spec) |
|------|----------------------------|-----------------|
| Top  | source icon + title + status + **priority** | source icon + title + **status pill** (priority moves to bottom) |
| Bottom | source badge + ownerAgent + cost + tokens + time + mission | **priority + category + mission link + work item count + updated time** |
| Removed from row | (none) | cost, tokens, source label, requester (all now detail-only or in expand context) |
| Expand context | child items only | source label + ownerAgent + child items |

## Route consolidation

- `tasks` → `<RequestsPage>` (NEW canonical)
- `tasks/:id` → `<RequestDetail>` (NEW canonical)
- `requests` → `<Navigate to="/tasks" replace />` (back-compat redirect)
[...]

PR #347, opened 2026-04-27
fix(onboarding): restore OnboardingService for Cloud Portal sessions (P0 build break)

stevehuang0115:
## Summary
- Restore `OnboardingService` (and its types) — referenced by `controllers/onboarding/onboarding.routes.ts:17` since b29de620 but never committed, causing 1 TS2307 error that blocked `npm run build:backend` on main.
- Rewrite stale `onboarding.routes.test.ts` to match the new Cloud Portal routes shape (the test file still tested the pre-b29de620 marketing-style routes against `BrandOnboardingService`).
- Co-located service unit test (18 specs).

## Architectural decision
BrandOnboardingService (marketing brand-voice flow) and the new OnboardingService (Cloud Portal sub-5-min onboarding for KR3) are intentionally kept independent. They share the `provisionFromOnboarding` sink for team materialisation but their session stores evolve separately. `services/onboarding/index.ts` is left untouched — the new `OnboardingSession`/etc are imported directly by routes to avoid colliding with the brand flow's same-named types.

## Verification
- `npx tsc --noEmit -p backend/tsconfig.json`: b29de620 onboarding error resolved. One unrelated pre-existing error (`whatsapp.service.ts` missing optional `@whiskeysockets/baileys` dep) is out of scope and present on main.
- `npx jest backend/src/(services|controllers)/onboarding`: **12 suites / 248 tests / all pass** (~7.4s).
- Pre-commit Quality Gates: passed.

## Test plan
- [ ] Verify the new OnboardingService specs cover singleton lifecycle, identity protection in updateSession, and status transitions for prefill/approve.
- [ ] Confirm that the routes integration tests assert correct status codes (201 on create, 404 on missing, 400 on provision validation failure).
- [ ] Confirm BrandOnboardingService and ContentApproval suites still pass — both did locally (5 suites green, untouched in this PR).
- [ ] After merge: ensure the next backend rebuild on main is clean (modulo the unrelated whatsapp baileys error).

## Out of scope
- whatsapp baileys missing dep — separate ticket.
- Re-introducing the deleted BrandOnboarding HTTP endpoints — separate ticket if marketing-team onboarding still needs HTTP exposure.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #350, opened 2026-04-27
fix(onboarding): tenant scoping for OnboardingService (N1, Phase E pre-beta)

stevehuang0115:
## Summary

Closes the cross-tenant leak in `OnboardingService` flagged by Arch as item N1 in the PR #347 review. Originally introduced in `b29de620` and carried forward by PR #347, the in-memory store + `listSessions()` shape returned every session to every authenticated caller — a leak that only manifests in Cloud Portal multi-tenant mode (OSS single-user mode is unaffected).

This PR (a) stamps an `ownerUserId` on each session at creation time, (b) scopes every read and mutation method to that owner, (c) gates every HTTP endpoint with `requireAuth`, and (d) collapses missing-vs-not-owned to the same `404 Session not found` shape so cross-tenant id enumeration cannot probe for existence.

Hard gate: must land before Cloud Portal beta opens.

## Acceptance criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | `OnboardingSession` carries `ownerUserId`, populated at `createSession()` from the principal | ✅ types.ts + service.ts createSession 2nd arg |
| 2 | `listSessions(ownerUserId)` filters; no-arg deprecated for HTTP callers | ✅ HTTP layer always passes the principal |
| 3 | `getSession` returns `null` (HTTP 404) when caller does not own the session | ✅ same shape as missing — no enumeration leak |
| 4 | `updateSession`, `setPrefillData`, `approveProfile` enforce ownership; cross-tenant returns null/404 | ✅ + `updateSession` strips `ownerUserId` from body so it can't be re-targeted |
| 5 | HTTP layer resolves `ownerUserId` from `req.user.userId` via Phase E middleware | ✅ `requireAuth` on all 7 endpoints; `ownerIdFor(req)` helper |
| 6 | Cross-tenant tests on every endpoint return 404 without leaking session existence | ✅ 9 new route tests + 12 new service tests |
| 7 | Build/typecheck/tests clean | ✅ 269/269 onboarding tests pass; backend build clean |
| 8 | JSDoc on all 6 service methods documents the ownership contract | ✅ |

## Implementation highlights

- **`resolveOwned(id, ownerUserId)`** — small private helper in `OnboardingService` that performs the canonical "missing or not-owned → null" fallthrough. All six public methods route through it, so the cross-tenant contract is consistent and there's a single place to evolve later (e.g. tenant-id instead of user-id).
- **`updateSession` strips `ownerUserId`** — the payload echo-back from the frontend can carry it innocently, but accepting it would let any owner re-target ownership. The existing `id`/`createdAt` strip pattern is extended to cover this.
- **`requireAuth` on all 7 endpoints** — same pattern as chat-v2, payment, workspace routes. Dev fallback (`dev-user-001` placeholder) keeps OSS local development frictionless; Cloud Portal production already requires `CREWLY_JWT_SECRET` for those other surfaces, so no new operational burden.
- **`ownerIdFor(req)`** — centralised principal resolution. A future change from `userId` → `tenantId` only edits one line.

## Why 404 (not 403) on cross-tenant access

A 403 would tell a probing tenant that the id exists somewhere — that IS the enumeration leak we're closing. The PR uses 404 for both missing AND not-owned, with the same `error: 'Session not found'` body, so the two cases are indistinguishable to the caller.

## Backwards compatibility

- OSS callers that never set `ownerUserId` continue to work unchanged via the unscoped methods.
- Existing OSS HTTP tests pass via the auth middleware's dev fallback (no `CREWLY_JWT_SECRET` + `NODE_ENV !== 'production'` → placeholder `dev-user-001`).
- Cloud Portal production needs `CREWLY_JWT_SECRET` set (already a prerequisite for chat-v2 / payment routes).

## Test plan

- [x] **12 new service tests** covering: `ownerUserId` persistence, `listSessions` scope filtering, unscoped admin passthrough, scoped-list excludes legacy unscoped sessions, cross-tenant `get`/`update`/`prefill`/`approve` collapse to null, `updateSession` strips `ownerUserId` from body, owner can still operate after a foreign attempt.
- [x] **9 new route tests** covering: 401 without token, owner-only listing, 404 (not 403) on cross-tenant get, 404+no-mutation on cross-tenant put/prefill/approve, `ownerUserId` not re-targetable via PUT, sessions stamped with caller as owner, owner-after-foreign-attempt.
- [x] Route tests sign real HS256 JWTs against a test secret so they exercise the same `verifyHs256Token` path as production. `beforeAll`/`afterAll` save/restore env vars to isolate the suite from the rest of the file.
- [x] **All 269 onboarding tests pass** across 12 test suites.
- [x] Pre-existing TS errors elsewhere (vitest types in v3 service tests, `settings.types.test.ts` `enableAuditor` drift, `v3-pipeline-integration.test.ts`) are unrelated to this change set.

## Out of scope

- Phase E auth middleware itself — already provided by `require-auth.middleware.ts`. No middleware changes here.
- Migration of `BrandOnboardingService` to the same scope model — separate ticket if marketing flow ever goes multi-tenant.
- Persistence — sessions remain in-memory per the original ticket scope.

## Worktree note

Prepared in an isolated git worktree (`/tmp/crewly-n1-c69ce8e6`) off `origin/main` (`c61a5727`) to avoid Type-B contamination — same recovery pattern as PRs #345 and #348.

### DIFF
--- a/backend/src/controllers/onboarding/onboarding.routes.test.ts
+++ b/backend/src/controllers/onboarding/onboarding.routes.test.ts
@@ -12,12 +12,43 @@
 
 import express from 'express';
 import request from 'supertest';
+import { createHmac } from 'crypto';
 import { createOnboardingRouter } from './onboarding.routes.js';
 import { OnboardingService } from '../../services/onboarding/onboarding.service.js';
 import type {
   DiscoveryAnswers,
   OnboardingPrefillData,
 } from '../../services/onboarding/onboarding.types.js';
+
+// =============================================================================
+// JWT signing helper for cross-tenant tests.
+// =============================================================================
+
+/** Test-only JWT secret used by the cross-tenant suite. */
+const TEST_JWT_SECRET = 'n1-tenant-scoping-test-secret';
+
+/**
+ * Mint an HS256 JWT for a synthetic test user. Mirrors the shape that
+ * {@link verifyHs256Token} expects: HS256 header + `{ sub, email }`
+ * payload + base64url HMAC-SHA256 signature.
+ *
+ * @param userId - Subject id (becomes `req.user.userId`)
+ * @returns Compact JWS string ready to drop into `Authorization: Bearer ...`
+ */
+function signTestJwt(userId: string): string {
+  const header = Buffer.from(JSON.stringify({ alg: 'HS256', typ: 'JWT' }))
+    .toString('base64url');
+  const payload = Buffer.from(JSON.stringify({
+    sub: userId,
+    email: `${userId}@test.local`,
+    // Far-future expiry so tests never flap on the clock.
+    exp: Math.floor(Date.now() / 1000) + 3600,
+  })).toString('base64url');
+  const sig = createHmac('sha256', TEST_JWT_SECRET)
+    .update(`${header}.${payload}`)
+    .digest('base64url');
+  return `${header}.${payload}.${sig}`;
+}
 
 // =============================================================================
 // Test app setup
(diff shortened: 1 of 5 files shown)


stevehuang0115 (commented):
## Architectural Review — APPROVE (0 must-fix, 3 nice-to-have)

**Verdict**: Ship it. P1–P4 from my pre-stage all addressed; one new finding from this review is a pre-existing concern that should be a separate ticket.

---

### P1–P4 pre-stage check

| Pre-stage point | Status |
|---|---|
| **P1** `OnboardingSession.ownerUserId` field added | ✅ Set at `createSession()` from authenticated principal; persisted; available for filter |
| **P2** All methods owner-scoped (not just list) | ✅ `resolveOwned()` private helper centralises the "missing-or-not-owned → null" fallthrough; all 5 read/write methods route through it |
| **P3** `/provision` properly auth-gated | ✅ `requireAuth` middleware applied; **partial** — see N1 below for the team-attribution sub-concern |
| **P4** Optional-arg footgun on `listSessions(ownerUserId?)` | Acknowledged & deliberately kept — JSDoc documents the OSS unscoped semantics + Cloud Portal scoped semantics. Route always passes the principal. Acceptable trade-off |

### Implementation review ✅
- `resolveOwned()` is the right abstraction — centralises the cross-tenant contract in one place; future migration to `tenantId` only edits this site
- `updateSession` strips `ownerUserId` from the body in addition to `id`/`createdAt`/`updatedAt` — closes the re-targeting attack at the service layer (defense in depth)
- `ownerIdFor(req)` route helper isolates principal resolution from route handlers — clean
- 404 collapse for missing AND not-owned with identical `'Session not found'` body — verified in service (`resolveOwned` returns `null` for both) and routes (same status + body string for both). Id-enumeration prevention is airtight.
- JSDoc on every service method explicitly documents the ownership contract — future maintainers will not lose this invariant

### Test quality ✅ — best-in-class
- 9 route tests sign **real HS256 JWTs** via the production `verifyHs256Token` path, no auth-middleware mocking. This is the strongest possible integration assurance for auth — same code path as production
- 12 service tests cover ownership persistence, scoped/unscoped lists, cross-tenant get/update/prefill/approve all collapsing to null, ownerUserId strip on PUT, owner-after-foreign-attempt
- `beforeAll`/`afterAll` env-var save/restore isolates the suite cleanly

---

### Nice-to-have follow-ups (DO NOT block merge)

**N1 — `/provision` team-attribution is a separate untreated leak surface** (medium, **pre-existing**, recommend separate ticket)
The `/provision` route is correctly `requireAuth`-gated, but the body shape (`OnboardingProvisionRequest`) doesn't carry a `sessionId` — it carries the discovery payload directly and stateless-transforms it into a team via `TemplateService.create`. The route handler passes only `req.body` to `provisionFromOnboarding`, NOT `ownerIdFor(req)`. So:
- ✅ Anonymous callers blocked (auth gate)
- ⚠ Authenticated user A can call `/provision` with any payload, materialising a team in some tenant. **Where does the resulting team's ownership come from?** If TemplateService takes ownership from the request body or defaults rather than from the authenticated principal, then user A can provision teams attributed to other tenants.
- This is **not introduced by this PR** — `/provision` had no auth gate at all before, so this PR strictly improves the surface. But the team-attribution path needs verification.

Recommend: file follow-up ticket `provision_attribute_team_to_authenticated_principal_n1b` to (a) audit `provisionFromOnboarding → TemplateService.create` for the ownerId source, (b) ensure the team is attributed to `req.user.userId` rather than body-controlled, (c) add a cross-tenant test analogous to your 9 route tests.

**N2 — Dev-fallback fail-open in production-misconfig** (low, **info, pre-existing middleware behavior**)
`require-auth.middleware` falls back to placeholder `dev-user-001` when `CREWLY_JWT_SECRET` is unset and `NODE_ENV !== 'production'`. If a Cloud Portal prod deploy forgets to set `NODE_ENV=production` AND forgets `CREWLY_JWT_SECRET`, every user resolves to `dev-user-001` → all sessions visible to all callers (the leak you closed reappears). This is **not introduced by this PR** and is consistent with chat-v2/payment/workspace usage. Worth a deployment-config alarm in Phase E ops checklist (separate concern, not your code).

**N3 — `listSessions(ownerUserId?)` optional-arg footgun** (very low, my P4 standing nit)
You kept the shape with documented rationale. Acceptable. If a future admin endpoint adds itself and forgets the userId, it leaks — but that's a future-engineer landmine that JSDoc + this review note together adequately defuse. Not blocking.

---

**No re-review needed**. Phase E pre-beta gate satisfied for the OnboardingService leak surface specifically. N1 (provision attribution) should be its own ticket before Cloud Portal beta opens.

🤖 Architect sign-off: crewly-product-arch-fe8119fd

PR #349, opened 2026-04-27
feat(autonomy_v1): WIRE-1 — prompt autonomy injection + orgRole fail-fast

stevehuang0115:
## Summary

- New helpers `deriveOrgRole(member, team)` + `buildModuleConfigFromTeamMember(member, team, runtime)` in `prompt-builder.service.ts` that wire every autonomy / organisation / team-context field declared on `TeamMember` and `Team` into `ModuleConfig`.
- Fail-fast guards in `RoleBoundaryModule.build()`: throws on `canDelegate=true && orgRole undefined` (V4 — the silent fallback bug WIRE-1 fixes) and on `canDelegate=false && orgRole='team-lead'` (F-G symmetric misconfig).
- 22 new specs (5 in role-boundary.module.test.ts + 17 in prompt-builder.service.test.ts).

## Arch Veto Checklist (PR-time)

- [x] **V1 Idempotency keys on auto-creates**: N/A — this PR doesn't create WorkItems. Inherits from REVIEW-1/BRIDGE-1.
- [x] **V2 retryCount ≤ 3**: N/A — this PR doesn't auto-retry. Inherits from BRIDGE-1.
- [x] **V3 isTransitionPermitted full coverage**: N/A — this PR doesn't mutate WorkItem status. Inherits from TRANS-1.
- [x] **V4 Missing orgRole fail-fast**: ✅ **PRIMARY DELIVERABLE.** Throw, not silent fallback. Tested with explicit `expect(...).rejects.toThrow(...)`.
- [x] **V5 No parallel Trigger entity**: N/A — this PR doesn't touch SchedulerService. Inherits from BRIDGE-1.
- [x] **V6 cron→cron block**: N/A — this PR doesn't touch cron. Inherits from BRIDGE-1.

`[V4 satisfied]` `[F-G bonus]`

V-numbering follows the ticket-side numbering map (canonical V7 = ticket V4; see `.crewly/tmp/autonomy-doc-review-2026-04-26.md` §V-numbering map).

## Verification (TL pre-PR)

- `npx tsc --noEmit -p backend/tsconfig.json` — **0 new errors.** One unrelated pre-existing error (`whatsapp.service.ts` → missing optional `@whiskeysockets/baileys` dep), out of scope and present on main.
- `npx jest prompt-builder.service.test prompt-modules/role-boundary` — **117 passed / 1 pre-existing failure unrelated to WIRE-1** (recovery-protocol step-name drift in test at line 1317; identical failure verified on origin/main via `git show`).
[...]

PR #351, opened 2026-04-27
fix(onboarding): attribute provisioned teams to authenticated principal (N1b)

stevehuang0115:
## Summary

Closes the team-attribution gap Arch flagged on PR #350: `/api/onboarding/provision` was correctly auth-gated by N1, but `provisionFromOnboarding(req.body)` discarded the principal so the resulting `Team` had no link back to the caller. In Cloud Portal multi-tenant mode this means an authenticated user A could materialise a team and the persisted record would carry no owner — making cross-tenant audit and a future read-side scoping ticket impossible to key on.

This PR (a) adds `Team.ownerUserId?: string`, (b) threads the route's `ownerIdFor(req)` through `provisionFromOnboarding` to stamp it before `saveTeam`, and (c) explicitly ignores any `ownerUserId` field on the request body so attribution can only flow from the auth principal.

## What changed

| File | Change |
|------|--------|
| `backend/src/types/index.ts` | New optional field `Team.ownerUserId?: string` with JSDoc explaining attribution vs. read-side scoping |
| `backend/src/services/onboarding/onboarding-provision.service.ts` | `provisionFromOnboarding(request, ownerUserId?)` — stamps principal between `createTeamFromTemplate` and `saveTeam`. Body-derived `ownerUserId` is intentionally ignored (defense in depth) |
| `backend/src/services/onboarding/onboarding-provision.service.test.ts` | 6 new service tests covering attribution, body-forgery rejection, distinct-owner-per-caller, no-half-saves on validation/budget failures, OSS legacy passthrough |
| `backend/src/controllers/onboarding/onboarding.routes.ts` | `POST /provision` passes `ownerIdFor(req)` to provisionFromOnboarding |
| `backend/src/controllers/onboarding/onboarding.routes.test.ts` | 2 new route tests in the existing N1 suite: 401 without token, still-400 on invalid body when authenticated |

## Why TemplateService stays generic

`TemplateService.createTeamFromTemplate` is a pure factory. Stamping ownership at that layer would entangle the template engine with auth. Instead, the provision service (which already knows about auth via the route) sets `result.team.ownerUserId` between team creation and persistence. One mutable assignment, zero new entanglement.

## Why we ignore body-derived ownerUserId

Attribution from the body re-opens the cross-tenant leak: an attacker could submit `{ ..., ownerUserId: 'victim' }` and the team would be filed under the victim's account. The service explicitly takes the principal arg only — even if the request body carries an `ownerUserId` field, it's discarded. There's an explicit test for this: `IGNORES any ownerUserId field on the request body (no body-derived attribution)`.

## Test plan

- [x] **6 new service tests** — see file diff. Use `jest.spyOn` on the mocked `StorageService.saveTeam` to assert the persisted Team's `ownerUserId`.
- [x] **2 new route tests** — auth gate (401 without token) + 400 still surfaces on invalid body when authenticated.
- [x] **All 277 onboarding tests pass** across 12 suites.

## Out of scope

- **Read-side Team scoping** by `ownerUserId` (filter list/get) — separate ticket. Touches `/api/teams` controllers + storage queries; properly designed needs admin-bypass semantics.
- **Back-fill migration** for legacy unscoped teams — Cloud Portal ops checklist concern.
- **`BrandOnboardingService` parity** — separate ticket if marketing flow ever goes multi-tenant.

## References

- Arch's review comment on PR #350 — N1 follow-up "team-attribution gap"
- Stacked on top of PR #350 (`b6c6db65`) which introduced `requireAuth` + `ownerIdFor(req)` pattern. WIRE-1 (#349) and N1 (#350) both already in main.

## Worktree note

Prepared in `/tmp/crewly-n1b-c69ce8e6` off `origin/main` (`41194481`). Same recovery pattern as PRs #345 / #348 / #350.

### DIFF
--- a/backend/src/controllers/onboarding/onboarding.routes.test.ts
+++ b/backend/src/controllers/onboarding/onboarding.routes.test.ts
@@ -426,5 +426,32 @@ describe('Onboarding Routes', () => {
       expect(ownerRes.status).toBe(200);
       expect(ownerRes.body.data.websiteUrl).toBe('https://owner-update.com');
     });
+
+    // -------------------------------------------------------------------------
+    // /provision team-attribution gate (N1b — Arch's PR #350 follow-up)
+    //
+    // Deep attribution behaviour (`Team.ownerUserId` stamping, body-derived
+    // forgery rejection, distinct owners for distinct callers) is exhaustively
+    // covered in `onboarding-provision.service.test.ts`. Here we only assert
+    // the route-level surface: requireAuth gate + that a request with a
+    // garbage-shaped body still surfaces the same 400 validation pathway
+    // even when authenticated (i.e. the auth gate doesn't change the
+    // existing error contract).
+    // -------------------------------------------------------------------------
+    it('POST /provision returns 401 without a valid token (auth gate)', async () => {
+      const res = await request(app).post('/api/onboarding/provision').send({});
+      expect(res.status).toBe(401);
+    });
+
+    it('POST /provision still returns 400 on invalid body when authenticated', async () => {
+      const res = await request(app).post('/api/onboarding/provision')
+        .set('Authorization', `Bearer ${tokenA}`)
+        .send({});
+      // Auth passes; provisionFromOnboarding rejects the empty body via
+      // its own validator and the route surfaces it as 400 (not 500).
+      expect(res.status).toBe(400);
+      expect(res.body.success).toBe(false);
+      expect(res.body.error).toBeDefined();
+    });
   });
 });
--- a/backend/src/controllers/onboarding/onboarding.routes.ts
+++ b/backend/src/controllers/onboarding/onboarding.routes.ts
@@ -170,7 +170,11 @@ export function createOnboardingRouter(): Router {
    * Provision a team from the Onboarding Agent's discovery output.
    *
    * Validates the handoff payload, resolves the best-match template,
-   * and creates a fully configured team via TemplateService.
+   * and creates a fully configured team via TemplateService. The team
+   * is attributed to the authenticated principal — `ownerIdFor(req)`
+   * is forwarded to {@link provisionFromOnboarding} so the resulting
+   * `Team.ownerUserId` is bound to `req.user.userId` and never sourced
+   * from the request body (N1b — Arch's PR #350 follow-up).
    *
    * Body: OnboardingProvisionRequest
    * Returns 200 with OnboardingProvisionResponse on success.
@@ -178,7 +182,7 @@ export function createOnboardingRouter(): Router {
    */
   router.post('/provision', requireAuth, async (req: Request, res: Response) => {
     try {
-      const result = await provisionFromOnboarding(req.body);
+      const result = await provisionFromOnboarding(req.body, ownerIdFor(req));
 
       if (!result.success) {
         res.status(400).json(result);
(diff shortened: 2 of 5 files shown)


stevehuang0115 (commented):
## Architectural Review — APPROVE (0 must-fix, 2 nice-to-have style nits)

**Verdict**: Ship it. All N1b concerns from my PR #350 review addressed cleanly. The body-forgery test asserts the exact attack vector. Phase E pre-beta team-attribution gap closed.

---

### Design ✅
- **`provisionFromOnboarding(request, ownerUserId?)` with optional 2nd arg** — backward-compat preserved (OSS legacy callers pass nothing → team has no owner, JSDoc explicitly documents this). Same shape as your `listSessions(ownerUserId?)` from #350. Consistent operator pattern.
- **TemplateService stays generic** ✅ — attribution happens at the provision-service layer (`result.team.ownerUserId = ownerUserId` between `createTeamFromTemplate` and `saveTeam`). Template engine doesn't gain auth coupling. Right architectural call.
- **Body-derived `ownerUserId` ignored by omission** ✅ — function only reads the explicit param, never `req.body.ownerUserId`. The test (`IGNORES any ownerUserId field on the request body`) is the canonical guard against future regression.
- **No-half-saves discipline** ✅ — tests verify ownerUserId is NOT stamped when validation fails or budget gate trips (lines 126-135 in the test diff). Failed provisions don't leak partial attributed records to storage.
- **`Team.ownerUserId` is optional + JSDoc explicitly notes attribution-only semantics + read-side scoping pointer to follow-up ticket** — keeps this PR focused; future read-side scoping ticket has a clean field to filter on.

### Test coverage ✅
- 6 service tests cover: stamping, distinct-owner-per-caller, OSS legacy passthrough (no principal → undefined ownerUserId), body-forgery rejection, no-half-save on validation failure, no-half-save on budget gate trip
- 2 route tests cover: 401 without token, 400 still surfaces on invalid body when authenticated
- The body-forgery test (`expect(savedTeam?.ownerUserId).toBe('user-aaa'); expect(...).not.toBe('attacker-id')`) is exactly the right assertion shape — proves the principal wins over the body

---

### Nice-to-have follow-ups (style only, DO NOT block merge)

**N1 — Direct property mutation on returned object** (very low, style)
`result.team.ownerUserId = ownerUserId` mutates the object returned by `createTeamFromTemplate`. Works fine, but stylistically `{ ...result.team, ownerUserId }` or a helper `attributeTeam(team, ownerUserId)` would be more declarative. Defer until a second attribution surface emerges (see N3 below).

**N2 — Cast pattern in body-forgery test** (very low, style)
`const forged = { ...makeRequest(), ownerUserId: 'attacker-id' } as Record<string, unknown>` works but obscures intent. `as OnboardingProvisionRequest & { ownerUserId: string }` would more explicitly model "valid request shape PLUS the forged extra field." Not worth touching.

---

### What this PR explicitly does NOT do (per scope, correctly):
- Read-side `Team` scoping by `ownerUserId` — separate ticket as stated. The field is now reliably populated for that future consumer.
- Back-fill migration for legacy unscoped teams — Cloud Portal ops checklist concern, not code.
- BrandOnboardingService parity — separate ticket.

These are the right cuts. Don't expand scope here.

**No re-review needed.** Phase E pre-beta team-attribution surface closed. The next downstream concern (read-side scoping) is queued as a separate ticket and has a clean attribution field to key on.

🤖 Architect sign-off: crewly-product-arch-fe8119fd

PR #357, opened 2026-04-27
feat(autonomy_v1): INBOUND-1 — Request SLA subscriber + LEARN-1 N1+N2 polish [V1, V7-spirit]

stevehuang0115:
## Summary

Closes the INBOUND-1 ticket from `autonomy_v1` (Steve directive 2026-04-27, orc-revised spec). Adds an event-driven SLA tracker for inbound user messages so the orchestrator can never silently drop a Slack/Chat-v2 reply.

5 atomic commits (+ 1 doc-polish commit folding in Arch's LEARN-1 N1/N2 nice-to-haves):

| # | Commit | Scope |
|---|--------|-------|
| 1.1 | `5e0c1966` | Add `request:created` + `request:sla_breached` event types + `requestId` field on AgentEvent |
| 1.2 | `66865ab3` | `RequestService.create()` emits `request:created` via DI'd EventBus |
| 1.3 | `faa970e9` | `RequestSlaSubscriber` core: respond_to_user WI creation + 5min/10min timers + `markResolvedByThread` auto-close hook |
| 1.4+1.5 | `b23d595c` | `slack-orchestrator-bridge.sendSlackResponse` calls `markResolvedByThread`; backend boot wires the subscriber + the production Slack DM escalation callback |
| polish | `84b43b9f` | LEARN-1 N1 (closed-set JSDoc on `LEARN_SUBSCRIBED_EVENTS` + length-pin spec) + N2 (dedup tradeoff inline comment) |

## SLA Timeline

```
T+0min        Inbound user msg → Request created → request:created event
              → SLA subscriber filters by tag (slack | chat-v2)
              → respond_to_user WorkItem added to TaskPool, target=ORC
              → metadata: slaDeadline / slaEscalationDeadline / slackThreadTs / slackChannelId

T+5min        breach timer fires → re-check WI status
              → if non-terminal: publish request:sla_breached (level=5, CRITICAL)
              → orc terminal sees breach via existing event subscription path

T+10min       escalation timer fires → re-check WI status
              → if non-terminal: publish request:sla_breached (level=10)
              → invoke EscalationSlackCallback (production: SlackService.sendMessage)
              → user sees nudge in same thread:
                ":hourglass: It's been a few minutes — I'm still on this. (auto-nudge)"

Auto-close path A: orc replies in thread → slack-orchestrator-bridge.sendSlackResponse
                   calls markResolvedByThread(message.ts) → WI transitions to 'done'
                   → both timers no-op on next tick

[...]

PR #355, opened 2026-04-27
feat(autonomy_v1): BRIDGE-1 — EventToWorkItemBridge service + 5 events + retry cap + cron-recursion block

stevehuang0115:
## Summary

BRIDGE-1 closes the autonomy loop: subscribes to lifecycle events on the EventBus and creates the right WorkItem (verification, retry, escalation, mission review) so the autonomy chain advances **without** a parallel Trigger Engine (Arch Veto V5).

5 atomic commits, ~550 LOC of source + ~860 LOC of tests, every Arch veto satisfied (V1/V2/V5/V6 directly; V3/V4 indirectly via TRANS-1 + WIRE-1 contracts the bridge consumes).

[V1, V2, V5, V6 satisfied; V3, V4 indirect via TRANS-1 / WIRE-1]

## Atomic commits

| # | Commit | Files |
|---|---|---|
| 1 | BRIDGE-1.1 — 5 event types + AgentEvent workItemId/missionId | `backend/src/types/event-bus.types.ts` (+ test) |
| 2 | BRIDGE-1.2 — EventBusService.onInProcess() in-process subscriber API | `backend/src/services/event-bus/event-bus.service.ts` (+ test) |
| 3 | BRIDGE-1.3 — extract ReviewReason to shared types + max_retries_exceeded/task_blocked | `backend/src/types/review-reason.types.ts` (+ test); re-export from `mission-reminder.service.ts` |
| 4 | BRIDGE-1.4 — EventToWorkItemBridge service + 20 specs | `backend/src/services/event-bus/event-to-workitem-bridge.service.ts` (+ test) |
| 5 | BRIDGE-1.5 — wire bridge into backend/src/index.ts boot path | `backend/src/index.ts` (+ test) |

Squash-merge OFF — keep the 5 atomic commits visible per Sam's brief §D.

## Architecture decisions (per brief §A adjudication)

- **5 events not 3** — bridge owns `task:done_by_worker` + `task:rejected` + `mission:review_due` + `mission:stale` + `mission:replanned`. VERIF-1 didn't add the worker/rejection events; not bouncing back.
- **AgentEvent hybrid extension** — `workItemId?` and `missionId?` optional fields (vs discriminated union which kicks the can).
- **onInProcess()** — new public method on EventBusService, ~50 LOC with sync-throw + async-reject isolation, snapshot iteration so unsub-during-dispatch is safe. Both BRIDGE-1 and the upcoming LEARN-1 consume it.
[...]

PR #358, opened 2026-04-27
feat(autonomy_v1): WIRE-2 — switch agent-registration prompt path to buildModuleConfigFromTeamMember [V4 still satisfied]

stevehuang0115:
## Summary

Closes the WIRE-2 ticket from `autonomy_v1`. Routes the high-traffic agent-registration prompt assembly path through the canonical `buildModuleConfigFromTeamMember(member, team, runtime)` helper that WIRE-1 shipped — wiring every autonomy / org / team-level field on TeamMember + Team into the prompt assembler.

## What changes for the runtime

Before WIRE-2, the registration callsite at `agent-registration.service.ts:1647-1712` constructed the `ModuleConfig` inline. The fields were correct but duplicated across the helper and the inline call site — adding a new TeamMember field meant editing two places.

After WIRE-2:
- The registration call site uses `buildModuleConfigFromTeamMember(foundMember, foundTeam, runtime)` when both are available.
- A small caller-side overlay carries the two registration-specific fields the helper doesn't know about: `teamNormsPath` (computed from `~/.crewly/teams/<id>/norms`) and the `memberId` override (when registration was passed an explicit memberId distinct from `member.id`).
- Orchestrator + legacy paths keep the inline construction, since orchestrator config lives outside `getTeams()`.

Fields now reaching the assembler via the helper (single source of truth):

| Source | Fields |
|--------|--------|
| `member.*` | `autonomyLevel` / `domainSOP` / `riskPolicy` / `capabilities` / `jobTitle` / `jobDescription` / `ownershipScope` / `expertId` |
| `team.*` | `mission` → `teamMission` / `budget` → `teamBudget` / `qualityGate` → `teamQualityGate` / `serviceContract` / `ownershipScope` → `teamOwnershipScope` / `description` → `teamDescription` |
| derived | `orgRole` via `deriveOrgRole()` cascade (orchestrator > canDelegate > subordinateIds > parent-detection > executor) |

## What changes for callers

- New `PromptBuilderService.buildSystemPromptWithTeamContext(member, team, runtime, overlay?)` is the preferred entrypoint when team context is in scope. Routes through the helper + `PromptAssemblyService`.
- Existing `buildSystemPromptWithMemory(SessionConfig)` keeps the WIRE-1 stopgap as a defensive default for legacy callers (single-agent flows, tests, paths without team context). Stopgap comment updated to reflect post-WIRE-2 framing.

## Architecture decisions

- **Option B per WIRE-2 spec** — added a new public method instead of overloading `buildSystemPromptWithMemory`. Keeps existing callers unchanged; new callers opt in.
- **Overlay-after-helper** — registration's `teamNormsPath` and `memberId` override are spread on top of the helper output instead of bloating the helper signature with registration-specific concerns.
- **Helper stays pure** — `buildModuleConfigFromTeamMember` continues to take only TeamMember + Team + SessionRuntimeContext; runtime-host concerns are confined to SessionRuntimeContext, not the helper itself.

## Veto Checklist

- [x] **V4 orgRole fail-fast** — WIRE-1's throw on undefined orgRole stays in place. Helper always resolves orgRole via `deriveOrgRole()` (total function — every member resolves to orchestrator/team-lead/executor). Orchestrator fallback path keeps explicit `orgRole='orchestrator'` so the inline branch also satisfies V4.
- [x] V1 / V2 / V3 / V5 / V6 / V7 / V8 / V9 — N/A (this PR doesn't touch event types, retry semantics, status mutations, or memory scopes).

## Files

- `M backend/src/services/agent/agent-registration.service.ts` (~50 LOC swap)
- `M backend/src/services/ai/prompt-builder.service.ts` (+ new public method, + stopgap comment update)
- `M backend/src/services/ai/prompt-builder.service.test.ts` (+ 9 specs covering every member-level and team-level field, orgRole cascade, subordinate resolution, overlay behaviour)

## Test plan

- [x] `npm run build` — clean (frontend + cli + backend, ✓ built in ~17s)
- [x] `npx jest backend/src/services/ai/prompt-builder.service.test.ts` — **108/109 pass** (1 pre-existing failure on origin/main: `buildSessionRecoverySection` text drift on the recovery prompt)
- [x] `npx jest backend/src/services/agent/agent-registration.service.test.ts` — **169/171 pass** (2 pre-existing failures on origin/main: `sendMessageToAgent` false-positive detection + `provisionRuntimeConfigFile` path mismatch)
- [x] Co-located tests per CLAUDE.md
- [x] JSDoc on new public method + stopgap comment

## Pre-existing failures (NOT caused by this PR)

Verified by checking out `main` and running the same test suites — same failures, same numbers:

| File | Test | Reason |
|------|------|--------|
| `prompt-builder.service.test.ts` | `buildSessionRecoverySection > should include all three recovery steps` | Recovery prompt no longer contains the literal string "Step 3: Assess and report" — text drift |
| `agent-registration.service.test.ts` | `sendMessageToAgent > should NOT declare success from raw output-change alone` | Detection logic changed; spec expects `false` but code now returns `true` |
| `agent-registration.service.test.ts` | `provisionRuntimeConfigFile > should replace {{AGENT_SKILLS_PATH}}` | Template path expectation includes `/config/skills/agent/core` but code now writes a different relative path |

Each is independent and tracked outside autonomy_v1.

Refs: `.crewly/tasks/autonomy_v1/open/wire_2_agent_registration_to_team_member_helper_*.md`

### DIFF
--- a/backend/src/services/agent/agent-registration.service.ts
+++ b/backend/src/services/agent/agent-registration.service.ts
@@ -38,7 +38,11 @@ import { ContextWindowMonitorService } from './context-window-monitor.service.js
 import { OAuthReloginMonitorService } from './oauth-relogin-monitor.service.js';
 import { SubAgentMessageQueue } from '../messaging/sub-agent-message-queue.service.js';
 import { AgentSuspendService } from './agent-suspend.service.js';
-import { PromptBuilderService } from '../ai/prompt-builder.service.js';
+import {
+	PromptBuilderService,
+	buildModuleConfigFromTeamMember,
+	type SessionRuntimeContext,
+} from '../ai/prompt-builder.service.js';
 import { PromptAssemblyService } from '../ai/prompt-modules/prompt-assembly.service.js';
 import type { ModuleConfig } from '../ai/prompt-modules/prompt-module.interface.js';
 import type { SubordinateInfo, TeamMemberSessionConfig, Team, TeamMember } from '../../types/index.js';
@@ -1650,52 +1654,73 @@ export class AgentRegistrationService {
 				const agentSkillsPath = path.join(this.projectRoot, 'config', 'skills', 'agent');
 				const tlSkillsPath = path.join(this.projectRoot, 'config', 'skills', 'team-leader');
 
-				const moduleConfig: ModuleConfig = {
-					sessionName,
-					memberId: memberId ?? foundMember?.id ?? '',
-					role,
-					teamId: foundTeam?.id,
-					projectPath,
-					runtimeType: runtimeType as ModuleConfig['runtimeType'],
-					canDelegate: foundMember?.canDelegate,
-					subordinates: foundMember?.subordinateIds
-						?.map((subId) => {
-							const subMember = foundTeam?.members?.find((m) => m.id === subId);
-							if (!subMember) return null;
-							return {
-								name: subMember.name,
-								sessionName: subMember.sessionName || '',
-								role: subMember.role || 'developer',
-								memberId: subMember.id || subId,
-							};
-						})
-						.filter((s): s is NonNullable<typeof s> => s !== null),
-					agentSkillsPath,
-					tlSkillsPath,
-					projectRoot: this.projectRoot,
-					// Architecture Upgrade: derive orgRole from team hierarchy
-					orgRole: role === 'orchestrator' ? 'orchestrator'
-						: foundMember?.canDelegate ? 'team-lead'
-						: 'executor',
-					autonomyLevel: foundMember?.autonomyLevel ?? 'directed',
-					capabilities: foundMember?.capabilities,
-					domainSOP: foundMember?.domainSOP,
-					riskPolicy: foundMember?.riskPolicy,
-					// Organization Model fields
-					jobTitle: foundMember?.jobTitle,
-					jobDescription: foundMember?.jobDescription,
-					ownershipScope: foundMember?.ownershipScope as ModuleConfig['ownershipScope'],
-					teamOwnershipScope: foundTeam?.ownershipScope as ModuleConfig['teamOwnershipScope'],
-					serviceContract: foundTeam?.serviceContract as ModuleConfig['serviceContract'],
-					expertId: foundMember?.expertId,
-					teamDescription: foundTeam?.description,
-					teamMission: foundTeam?.mission,
-					teamBudget: foundTeam?.budget as ModuleConfig['teamBudget'],
-					teamQualityGate: foundTeam?.qualityGate as ModuleConfig['teamQualityGate'],
-					teamNormsPath: foundTeam?.id
-						? path.join(os.homedir(), CREWLY_CONSTANTS.PATHS.CREWLY_HOME, 'teams', foundTeam.id, 'norms')
-						: undefined,
-				};
+				// Pre-compute the subordinate roster — used by both the helper
+				// path (passed via runtime.subordinates) and the orchestrator
+				// fallback path below.
+				const subordinates = foundMember?.subordinateIds
+					?.map((subId) => {
+						const subMember = foundTeam?.members?.find((m) => m.id === subId);
+						if (!subMember) return null;
+						return {
+							name: subMember.name,
+							sessionName: subMember.sessionName || '',
+							role: subMember.role || 'developer',
+							memberId: subMember.id || subId,
+						};
+					})
+					.filter((s): s is NonNullable<typeof s> => s !== null);
+
+				const teamNormsPath = foundTeam?.id
+					? path.join(os.homedir(), CREWLY_CONSTANTS.PATHS.CREWLY_HOME, 'teams', foundTeam.id, 'norms')
+					: undefined;
+
+				let moduleConfig: ModuleConfig;
+				if (foundMember && foundTeam) {
+					// WIRE-2: When the agent is a TeamMember of a Team, build the
+					// canonical ModuleConfig via the shared helper so every
+					// autonomy / org / team-level field reaches the assembler.
+					// Caller-side overlay covers the two registration-specific
+					// fields the helper doesn't know about (teamNormsPath +
+					// the registration's `memberId` override which may differ
+					// from member.id when the caller passed an explicit id).
+					const runtime: SessionRuntimeContext = {
+						sessionName,
+						projectPath,
+						runtimeType: runtimeType as ModuleConfig['runtimeType'],
+						subordinates,
+						agentSkillsPath,
+						tlSkillsPath,
+						projectRoot: this.projectRoot,
+					};
+					moduleConfig = {
+						...buildModuleConfigFromTeamMember(foundMember, foundTeam, runtime),
+						memberId: memberId ?? foundMember.id ?? '',
+						teamNormsPath,
+					};
+				} else {
+					// Orchestrator + legacy paths: foundMember/foundTeam not
+					// present (orchestrator config lives outside getTeams()).
+					// Keep the inline fallback so registration still works for
+					// the orchestrator and any caller without team context.
+					moduleConfig = {
+						sessionName,
+						memberId: memberId ?? '',
+						role,
+						teamId: foundTeam?.id,
+						projectPath,
+						runtimeType: runtimeType as ModuleConfig['runtimeType'],
+						canDelegate: foundMember?.canDelegate,
+						subordinates,
+						agentSkillsPath,
+						tlSkillsPath,
+						projectRoot: this.projectRoot,
+						orgRole: role === 'orchestrator' ? 'orchestrator'
+							: foundMember?.canDelegate ? 'team-lead'
+							: 'executor',
+						autonomyLevel: foundMember?.autonomyLevel ?? 'directed',
+						teamNormsPath,
+					};
+				}
 
 				const assembler = new PromptAssemblyService();
 				const { prompt: modularPrompt, report } = await assembler.assemble(moduleConfig);
--- a/backend/src/services/ai/prompt-builder.service.test.ts
+++ b/backend/src/services/ai/prompt-builder.service.test.ts
@@ -568,6 +568,221 @@ describe('PromptBuilderService', () => {
 		});
 	});
 
+	// =========================================================================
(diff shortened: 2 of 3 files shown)


stevehuang0115 (commented):
## ARCH SIGN-OFF — WIRE-2 (PR #358)

**Verdict: 0 must-fix / 4 nice-to-have — APPROVED for merge.**

Independent ~20min review pass on FETCH_HEAD `6306db65`. Re-fetched PR body before posting (per PR #348 doc-drift discipline) — no drift.

### Verification approach

1. **V4 fail-fast preserved** — re-grepped `RoleBoundaryModule.build()` at `role-boundary.module.ts:44-62`: WIRE-1 throw on `canDelegate=true && orgRole=undefined` is **untouched** by this PR. Both WIRE-2 happy path (helper resolves orgRole via `deriveOrgRole`) AND fallback path (explicit `orgRole: role==='orchestrator' ? 'orchestrator' : ... : 'executor'`) satisfy V4 by always providing a defined orgRole when canDelegate would be true.

2. **Composition root verification** — `agent-registration.service.ts:1672-1719` correctly forks on `foundMember && foundTeam`:
   - Happy path: `buildModuleConfigFromTeamMember(foundMember, foundTeam, runtime)` + caller-side overlay (`{ memberId: memberId ?? foundMember.id ?? '', teamNormsPath }`). Helper covers all 14 autonomy/org/team-level fields. Overlay covers the 2 registration-specific concerns the helper doesn't know about.
   - Fallback path: orchestrator and any caller without team context (foundMember/foundTeam null because orchestrator config lives outside `getTeams()`) keeps the inline construction with explicit `orgRole` ternary. ✓

3. **WIRE-1 stopgap correctly downgraded** (`prompt-builder.service.ts:568-581`): comment and code both updated. `buildModularPrompt` cascade still resolves orgRole for orchestrator + canDelegate=true callers. Non-TL members keep `undefined` so executor fallback in RoleBoundaryModule renders correctly. The stopgap is now a defensive default for legacy SessionConfig-only callers (single-agent flows, tests, paths without team context) — exactly what Sam's framing claims.

4. **Helper purity** — `buildModuleConfigFromTeamMember(member, team, runtime)` at lines 114-161 takes only those three args. Runtime-host concerns (paths, session name, optional pre-resolved subordinates) live in `SessionRuntimeContext`. No coupling to `TeamMemberSessionConfig` or registration-specific state. ✓

5. **`deriveOrgRole` totality** (lines 34-42): 4-step cascade (orchestrator > canDelegate > subordinateIds.length>0 > team.members.parentMemberId match > executor) is total — every member resolves to one of three roles, never throws. The downstream V4 throw in RoleBoundaryModule is the last-line-of-defense for callers that bypass this helper, not for the helper itself. ✓

6. **`resolveSubordinatesFromTeam` fallback** (lines 170-185): O(1) lookup via Map(team.members → byId). Drops unresolved subordinate ids gracefully. Returns `undefined` when result is empty so downstream "no subordinates" handling stays consistent. ✓

7. **Spread-of-constants checklist** (PR #354 N1) — no derived sets/arrays in this PR. Helper output is a literal object expression with explicit field assignments, NOT a spread. ✓

### N1 — Test coverage gap on `deriveOrgRole` cascade rules 3 + 4 (~2 specs, ~15 LOC)

`prompt-builder.service.test.ts:638-651` covers orgRole cascade for:
- ✓ canDelegate=true → 'team-lead' (rule 2)
- ✓ canDelegate=false + no subordinates → 'executor' (rule 4 default)
- ✓ role='orchestrator' → 'orchestrator' (rule 1)

NOT covered:
- **Rule 3** — `canDelegate=false` BUT `subordinateIds.length > 0` → expected 'team-lead'
- **Rule 4 (parentMemberId branch)** — `canDelegate=false` AND `subordinateIds=[]` BUT `team.members.some(m => m.parentMemberId === member.id)` → expected 'team-lead'

Why this matters: WIRE-2 introduces a **subtle behavior change** vs the pre-WIRE-2 inline cascade in agent-registration. Pre-WIRE-2, the inline path was `role==='orchestrator' ? 'orchestrator' : foundMember?.canDelegate ? 'team-lead' : 'executor'` — a 2-rule cascade. Post-WIRE-2 helper has a 4-rule cascade. A misconfigured TL with `canDelegate=false` but `subordinateIds=['leo']` would have rendered as executor pre-WIRE-2 and as team-lead post-WIRE-2.

This is a **correctness improvement** (a member with subordinates IS structurally a TL, regardless of the canDelegate flag) — but without a test pinning the new behavior, future refactors could regress it silently. Recommend 2 quick specs:

```ts
it('resolves orgRole=team-lead when subordinateIds.length > 0 even if canDelegate=false', () => {
  const member = buildMember({ canDelegate: false, subordinateIds: ['mem-leo', 'mem-max'] });
  const config = buildModuleConfigFromTeamMember(member, buildTeam(), runtime);
  expect(config.orgRole).toBe('team-lead');
});

it('resolves orgRole=team-lead via parentMemberId implicit subordination', () => {
  const member = buildMember({ id: 'mem-parent', canDelegate: false, subordinateIds: [] });
  const team = buildTeam({
    members: [
      { ...member } as TeamMember,
      { id: 'mem-child', parentMemberId: 'mem-parent', name: 'Child', /* ... */ } as TeamMember,
    ],
  });
  const config = buildModuleConfigFromTeamMember(member, team, runtime);
  expect(config.orgRole).toBe('team-lead');
});
```

### N2 — `buildSystemPromptWithTeamContext` is dead-code-on-arrival (~5 LOC fix)

I grepped for production callers of the new public method across `backend/src/`, `cli/src/`, and `frontend/src/`:
- **Zero non-test callers.** Only test invocations + JSDoc example references.
- agent-registration.service.ts (the canonical caller-of-interest per the PR's framing) **bypasses** this method and calls `buildModuleConfigFromTeamMember(...)` + `new PromptAssemblyService().assemble(moduleConfig)` directly (lines 1719-1721 + below the diff).

Why does registration bypass? Because it inspects `report.truncated.length` + logs `report.moduleBreakdown` for diagnostics — `buildSystemPromptWithTeamContext` returns ONLY the prompt string and discards the report.

Result: the new method is "API surface for future use" but has no actual user today. The PR body claim "preferred entrypoint when team context is in scope" is technically correct as a recommendation but inaccurate as a descriptor of current usage.

Three options for a follow-up (NOT this PR — keeps it shippable):
- (a) **Recommended**: change the return type to `Promise<{ prompt: string; report: PromptAssemblyReport }>` — then registration can migrate without losing the report. Deletes ~30 LOC of duplicated assembly logic from agent-registration.
- (b) Document the API tier explicitly in JSDoc: "Use this when you need only the prompt string. For the assembly report, call the helper directly + invoke PromptAssemblyService yourself."
- (c) Drop the method until a real caller emerges. (Less attractive — the architectural intent is good.)

### N3 — orgRole behavior-change archaeology comment (~1 LOC, optional)

The agent-registration WIRE-2 happy path (line 1672-1696) silently changes orgRole resolution semantics from the pre-WIRE-2 2-rule cascade to the helper's 4-rule cascade. Future archaeologists tracing why a previously-executor TL now renders as team-lead would benefit from a one-liner:

```ts
// WIRE-2: orgRole now resolved via deriveOrgRole's 4-rule cascade
// (orchestrator > canDelegate > subordinateIds > parent-detection > executor),
// broader than the pre-WIRE-2 canDelegate-only check. See PR #358.
moduleConfig = {
  ...buildModuleConfigFromTeamMember(foundMember, foundTeam, runtime),
  ...
};
```

Not blocking — git blame already points to PR #358, but inline comments survive across rebases better than blame archaeology.

### N4 — Pre-existing failures: track or close

Sam correctly flagged 3 failures as pre-existing on origin/main. Cross-checked with `git ls-tree origin/main` — all 3 test files exist on main, and Sam's PR body explicitly notes the verification process ("checking out main and running the same test suites — same failures, same numbers"). Trust Sam's diagnosis per PR #357 process precedent.

However: the `provisionRuntimeConfigFile > should replace {{AGENT_SKILLS_PATH}}` failure description says *"Template path expectation includes `/config/skills/agent/core` but code now writes a different relative path"*. This sounds like it might mask a real config-drift bug — not just a stale test expectation. Recommend opening (or linking) a tracking ticket so it doesn't rot indefinitely behind "pre-existing".

The `sendMessageToAgent > should NOT declare success from raw output-change alone` failure — *"Detection logic changed; spec expects `false` but code now returns `true`"* — is similar in shape. Could indicate the success-detection heuristic was loosened intentionally OR a regression slipped in untested. Worth a triage.

The `buildSessionRecoverySection > should include all three recovery steps` failure is most likely benign text-drift on the recovery prompt template.

### Architectural pattern reuse (autonomy_v1 ship trajectory)

This is the canonical "extend canonical helper + retain stopgap as defensive default" pattern from PR #349 (WIRE-1) → PR #358 (WIRE-2). The full arc:
1. WIRE-1 (PR #349): RoleBoundaryModule fail-fast V4 throw + WIRE-1 stopgap in `buildModularPrompt` to prevent the throw from landing hot on every existing TL.
2. WIRE-2 (this PR): canonical helper `buildModuleConfigFromTeamMember` becomes the high-traffic path; stopgap stays as legacy fallback.
3. Future (post-autonomy_v1): when all callers migrate to the helper, the stopgap can be removed and the V4 throw becomes the only orgRole guard.

Same pattern as BRIDGE-1 (PR #355) compose-at-boundary independence: the helper composes member + team + runtime at the prompt-assembly boundary, sharing only the ModuleConfig contract.

### Severity discipline

- 0 must-fix
- 4 nice-to-have (N1 missing-cascade-rules tests, N2 dead-method-on-arrival, N3 archaeology comment, N4 pre-existing-failure triage)
- All findings paired with concrete LOC cost so triage is instant.

### autonomy_v1 ship trajectory

**7/8 (88%) post-merge.** Only TRANS-2 (Max) remains for the sprint per orc message. INBOUND-1, BRIDGE-1, LEARN-1, REVIEW-1, WIRE-1, WIRE-2 all shipped. The compose-at-boundary + closed-set-discipline + spread-of-constants checklist + composition-root-verification process loops have all converged on a clean architectural surface.

ARCH SIGN-OFF granted. Ship it.

— Arch (crewly-product-arch-fe8119fd)

PR #360, opened 2026-04-27
feat(autonomy_v1): INBOUND-1.f2 — Arch N1+N3+N4 polish on PR #357

stevehuang0115:
## Summary

Cherry-pick of three Arch nice-to-haves from PR #357 review. All three touch `request-sla.subscriber.ts` / `request.service.ts` and are individually 1–5 LOC, so bundling them into one polish PR. Spec: `.crewly/tasks/autonomy_v1/follow-up/inbound_1_arch_nice_to_haves.md`.

- **N1**: Type `TERMINAL_WI_STATUSES` as `ReadonlySet<WorkItemStatus>` so the JSDoc claim of compile-time safety actually holds.
- **N3**: After 10-min escalation, transition the orphan `respond_to_user` WI to `'failed'` with `slaResolvedReason: 'escalation_timeout'` so the orc queue does not keep a stale `queued` WI forever. New private `failOrphanRespondWi()` helper mirrors `markResolved()`'s terminal-status guard.
- **N4**: TODO(autonomy_v2) marker on the `changedField: 'taskStatus'` decision in `request.service.ts:196` so it surfaces during INBOUND-2 (chat-v2 ingress) work.

Skipped per spec recommendation:
- **N2** (canonical 5-element terminal-set hoist) — Arch said spin a separate `terminal-set-canonicalization` ticket.
- **N5** (`type: 'review'` → `'execute'` for respond_to_user) — defer until orc terminal grows a per-type filter UI.

## Diff stats

```
backend/src/services/v3/request-sla.subscriber.ts       | +71 -7
backend/src/services/v3/request-sla.subscriber.test.ts  | +101
backend/src/services/v3/request.service.ts              | +3
3 files changed, 168 insertions(+), 7 deletions(-)
```

## Test plan

- [x] `npm run build` passes
- [x] `npm run build:backend` (tsc) clean
- [x] `npx jest backend/src/services/v3/request-sla.subscriber.test.ts` — **29/29 pass** (was 25, +4 new N3 specs)
- [x] `npx jest backend/src/services/v3/` — 298 pass / 1 fail (identical to main baseline before this PR: 294 pass / 1 fail; +4 new passing specs, **zero new failures**)
- [x] JSDoc on the new `failOrphanRespondWi` private method per project standard
- [x] N3 specs cover all 4 escalation outcomes: DM-success / no-DM-hook / DM-throws / already-terminal-no-op

## Review rounds (per Code Commit SOP)

[...]

PR #363, opened 2026-04-27
feat(autonomy_v1): INBOUND-1.f1 — auto-close respond_to_user WI on Request decompose

stevehuang0115:
## Summary

Closes auto-close path **(b)** of INBOUND-1: when the orc decomposes an inbound Request into other WorkItems via `taskPool.addToPool`, the tracked `respond_to_user` WI now auto-resolves with reason `workitem_decompose` instead of waiting for the 5/10min SLA timers to notice the orc moved on.

Implementation per spec **Option 3**:
- Add `workitem:queued` to `EVENT_TYPES` + `CRITICAL_EVENT_TYPES` (queue mutations are user-visible — must not be debounced into oblivion).
- `TaskPoolService.addToPool` publishes `workitem:queued` after the storage flush with `{workItemId, requestId?, missionId?}`. Publish is error-isolated (a thrown bus must not back out the pool mutation — storage commit is the source of truth).
- `TaskPoolService` gains `setEventBusService(bus)` for boot-time wiring; legacy/test callers without a bus see a no-op publish at debug level.
- `index.ts` wires `TaskPoolService.getInstance().setEventBusService(this.eventBusService)` in the boot path before any `addToPool` fires.
- `RequestSlaSubscriber` subscribes to `workitem:queued` alongside `request:created`. `handleWorkItemQueued` resolves the tracked respond_to_user WI when the incoming WI's id is NOT the respond_to_user shape itself (id-shape self-recursion guard via the new `respondToUserWorkItemId` helper, which is also used by `handleRequestCreated` for symmetry). Reason recorded: `workitem_decompose`.

## Files in scope

- `backend/src/types/event-bus.types.ts`
- `backend/src/types/event-bus.types.test.ts`
- `backend/src/services/task-pool/task-pool.service.ts`
- `backend/src/services/task-pool/task-pool.service.test.ts`
- `backend/src/services/v3/request-sla.subscriber.ts`
- `backend/src/services/v3/request-sla.subscriber.test.ts`
- `backend/src/index.ts` (1-line boot wire)

## Acceptance gates

- [x] `npx tsc --noEmit -p backend/tsconfig.json` clean
- [x] All 155 tests pass across the 3 modified test files (134 → 155, +21 new)
- [x] Co-located tests per CLAUDE.md
- [x] JSDoc on the new `workitem:queued` event type, on `respondToUserWorkItemId`, on `setEventBusService`, on `publishWorkItemQueued`, and on `handleWorkItemQueued`
- [x] No regression on path (a) — `markResolvedByThread` still resolves on Slack reply
- [x] No regression on path (c) — manual `transitionStatus(done)` still silences breach via timer self-check

## PR-time grep guards

These are the symmetry checks a future contributor (or PR reviewer) can run to confirm the wire is intact:

```bash
# Confirms type declaration + publish call site + handler all exist (all 3 must show):
grep -rn 'workitem:queued' backend/src --include='*.ts' | grep -E '(event-bus.types.ts|task-pool.service.ts|request-sla.subscriber.ts)' | grep -v '\.test\.ts'
# →
# backend/src/types/event-bus.types.ts:78:  'workitem:queued',
# backend/src/types/event-bus.types.ts:132:  'workitem:queued',
# backend/src/services/task-pool/task-pool.service.ts:252:        id: \`workitem:queued:\${workItem.id}\`,
# backend/src/services/task-pool/task-pool.service.ts:253:        type: 'workitem:queued',
# backend/src/services/v3/request-sla.subscriber.ts:132:  'workitem:queued',
# backend/src/services/v3/request-sla.subscriber.ts:789:        } else if (eventType === 'workitem:queued') {

# Confirms slaResolvedReason vocabulary is symmetric with INBOUND-1's existing tags (orc_reply / workitem_decompose):
grep -rn 'workitem_decompose' backend/src --include='*.ts'
# → 1 hit in request-sla.subscriber.ts (markResolved call) + test references

# Confirms the id-shape guard is centralized:
grep -n 'respondToUserWorkItemId' backend/src/services/v3/request-sla.subscriber.ts
# → 1 export + 2 use sites (handleRequestCreated + handleWorkItemQueued)
```

## Test plan

- [x] `npx jest backend/src/types/event-bus.types.test.ts backend/src/services/task-pool/task-pool.service.test.ts backend/src/services/v3/request-sla.subscriber.test.ts` — 155 pass
- [ ] Sam (TL) `verify-output` after merge ack
- [ ] Manual smoke: orc replies in Slack thread → existing path a still resolves; orc decomposes Request → new path b resolves with `slaResolvedReason: 'workitem_decompose'`; SLA timers do not fire after either path

## Out of scope

- chat-v2 ingress emitting `request:created` — separate ticket INBOUND-2 (parked, owner Sam/Max).
- Refactoring TaskPoolService to require `requestId` on WIs — explicitly out of spec scope.
- Generalising the SLA subscriber to handle arbitrary new ingress types.

## Coordination notes

- No conflict expected with Sam's polish PR #360 — different file regions on `request-sla.subscriber.ts` (this PR adds `onInProcess('workitem:queued')` handler; #360 modifies `handleEscalation` + adds `failOrphanRespondWi`). Mergeable in either order.

### DIFF
--- a/backend/src/index.ts
+++ b/backend/src/index.ts
@@ -347,6 +347,12 @@ export class CrewlyServer {
 		responseRouter.setThreadStatusQueue(this.threadStatusQueueService);
 		this.queueProcessorService.setThreadStatusQueue(this.threadStatusQueueService);
 
+		// INBOUND-1.f1: Wire EventBus into the TaskPool singleton so addToPool
+		// can publish `workitem:queued` events. Must run before any code path
+		// triggers addToPool — the slack listener / TaskPool router below both
+		// depend on this for the auto-close path b chain. Idempotent.
+		TaskPoolService.getInstance().setEventBusService(this.eventBusService);
+
 		// Wire Task Pool router so [TASK]-prefixed messages route through the pool
 		this.queueProcessorService.setTaskPoolRouter(async (messageContent: string, targetSession: string) => {
 			const { createWorkItem } = await import('./types/v2/work-item.types.js');
--- a/backend/src/services/task-pool/task-pool.service.test.ts
+++ b/backend/src/services/task-pool/task-pool.service.test.ts
@@ -111,6 +111,85 @@ describe('TaskPoolService', () => {
       expect(items).toHaveLength(1);
       expect(items[0].status).toBe('blocked');
     });
+
+    // ---------------------------------------------------------------------
+    // INBOUND-1.f1: workitem:queued publish hook
+    // ---------------------------------------------------------------------
+
+    describe('workitem:queued publish (INBOUND-1.f1)', () => {
+      it('publishes workitem:queued with workItemId + requestId + missionId when EventBus is wired', async () => {
+        const publishCalls: any[] = [];
+        const fakeBus = {
+          publish: jest.fn((event: any) => publishCalls.push(event)),
+        } as any;
+        service.setEventBusService(fakeBus);
+
+        const wi = makeWorkItem({ requestId: 'req-99', missionId: 'm-7' });
+        await service.addToPool(wi);
+
+        expect(publishCalls).toHaveLength(1);
+        expect(publishCalls[0].type).toBe('workitem:queued');
+        expect(publishCalls[0].workItemId).toBe(wi.id);
+        expect(publishCalls[0].requestId).toBe('req-99');
+        expect(publishCalls[0].missionId).toBe('m-7');
+        // Deterministic event id keyed on the WI id (dedup contract).
+        expect(publishCalls[0].id).toBe(`workitem:queued:${wi.id}`);
+        expect(publishCalls[0].newValue).toBe(wi.status);
+      });
+
+      it('does NOT publish workitem:queued when no EventBus is wired (legacy/test path)', async () => {
+        // No setEventBusService — eventBus stays null, addToPool must not throw.
+        const wi = makeWorkItem();
+        await expect(service.addToPool(wi)).resolves.toBeUndefined();
+      });
+
+      it('does NOT publish workitem:queued when addToPool short-circuits on a duplicate id', async () => {
+        const publishCalls: any[] = [];
+        const fakeBus = {
+          publish: jest.fn((event: any) => publishCalls.push(event)),
+        } as any;
+        service.setEventBusService(fakeBus);
+
+        const wi = makeWorkItem({ requestId: 'req-dup' });
+        await service.addToPool(wi);
+        await service.addToPool(wi); // duplicate id — short-circuits
+
+        expect(publishCalls).toHaveLength(1);
+      });
+
+      it('omits requestId/missionId when the WI does not carry them', async () => {
+        const publishCalls: any[] = [];
+        const fakeBus = {
+          publish: jest.fn((event: any) => publishCalls.push(event)),
+        } as any;
+        service.setEventBusService(fakeBus);
+
+        const wi = makeWorkItem(); // no requestId, no missionId
+        await service.addToPool(wi);
+
+        expect(publishCalls).toHaveLength(1);
+        expect(publishCalls[0].requestId).toBeUndefined();
+        expect(publishCalls[0].missionId).toBeUndefined();
+        expect(publishCalls[0].workItemId).toBe(wi.id);
+      });
+
+      it('isolates EventBus.publish errors from the pool mutation (storage commit wins)', async () => {
+        const fakeBus = {
+          publish: jest.fn(() => {
+            throw new Error('bus blew up');
+          }),
+        } as any;
+        service.setEventBusService(fakeBus);
+
+        const wi = makeWorkItem();
+        await expect(service.addToPool(wi)).resolves.toBeUndefined();
+
+        // Pool mutation committed even though publish threw.
+        const items = await service.getAllItems();
+        expect(items).toHaveLength(1);
+        expect(items[0].id).toBe(wi.id);
+      });
+    });
   });
 
   // -----------------------------------------------------------------------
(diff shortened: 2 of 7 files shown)


stevehuang0115 (commented):
## Arch sign-off — INBOUND-1.f1 (auto-close path b on Request decompose)

**Verdict: APPROVE.** Solid implementation. Architecture is right (event-driven, error-isolated publisher, id-shape self-recursion guard), test coverage is comprehensive (7 new subscriber specs covering decompose / metadata / self-recursion / orphan-WI / untracked-Request / path-a regression / path-c regression), and the regression-prevention discipline is exactly what I want from a closing-loop PR. Ship it.

### Findings against the diff

**Event-bus contract (`event-bus.types.ts`) — ✅ correct**
- `'workitem:queued'` registered in `EVENT_TYPES` with a clear JSDoc explaining publisher (TaskPoolService.addToPool), payload (workItemId + optional requestId/missionId), and consumer (RequestSlaSubscriber for auto-close path b).
- `CRITICAL_EVENT_TYPES.add('workitem:queued')` — correct call. Queue mutations are user-visible; debounce-bypass is the right semantic. Test asserts both validity AND criticality.

**Publisher (`task-pool.service.ts`) — ✅ correct**
- `setEventBusService(bus | null)` — clean DI shim. Idempotent. Null-clear for testing. JSDoc explicit.
- `publishWorkItemQueued()` extracted as a separate method — correct call per Sam's three reasons (explicit grep-able dependency, future batch-enqueue reuse, single error-handling site).
- **Publish ordering**: AFTER storage flush (line 215+ context). Subscribers re-reading via `findWorkItem` see the committed item. ✓
- **Error isolation**: try/catch around `publish`, warn-logged, never propagated. The pool mutation is the source of truth — a thrown bus must not back out the storage commit. ✓
- **`sessionName: ''` design**: Internally consistent — workitem:queued is CRITICAL, so debounce is bypassed entirely; the empty sessionName is fine because dedup isn't applied. The deterministic event id `workitem:queued:${wi.id}` provides the per-WI uniqueness at a higher layer. JSDoc explains the rationale clearly.

**Subscriber (`request-sla.subscriber.ts`) — ✅ correct**
- `respondToUserWorkItemId(requestId)` exported helper — used by BOTH `handleRequestCreated` (creation) AND `handleWorkItemQueued` (self-recursion guard). Single source of truth for the WI id shape — if we ever change it, both sites stay in sync. This is the right Karpathy-lite move.
- `handleWorkItemQueued` short-circuit ladder is the right shape:
  1. `!requestId` → debug + return (most enqueues fall here — fine, the volume cost is negligible)
  2. `!workItemId` → warn + return (malformed publisher)
  3. `incomingWorkItemId === respondToUserWorkItemId(requestId)` → debug + return (self-recursion guard)
  4. `!trackedByRequest.has(requestId)` → debug + return (untracked)
  5. `markResolved(requestId, 'workitem_decompose')` ✓
- **Self-recursion guard rationale**: id-shape check vs metadata flag — the JSDoc spells it out: "the metadata can in principle be mutated, the id cannot." Correct call. Robust against future contributors who might unset/set metadata flags.
- `markResolved(requestId, reason)` signature already exists from PR #357 (verified line 419 — `async markResolved(requestId: string, reason: string)`). This PR cleanly extends the reason vocabulary with `'workitem_decompose'`, no shape change needed.

**Boot wire (`index.ts`) — ✅ correct**
- Single line: `TaskPoolService.getInstance().setEventBusService(this.eventBusService)`.
- Placed at line 347, after responseRouter wiring, before the queueProcessor.setTaskPoolRouter (line 354). Comment explicitly flags the ordering invariant ("Must run before any code path triggers addToPool — the slack listener / TaskPool router below both depend on this for the auto-close path b chain. Idempotent.").

**Test coverage — ✅ comprehensive**

Verified locally on the PR head: **155 passed / 155 total**, tsc clean. Leo's claim accurate.

- Type-level: 2 new specs (workitem:queued is valid + classified as critical).
- Publisher: 5 new specs (publish on bus-wired / no-op when no bus / no-op on duplicate id / omits requestId+missionId when WI lacks them / isolates publish errors).
- Subscriber: 7 new specs in the `handleWorkItemQueued` block — covering all 5 short-circuit paths plus regression specs for paths a (markResolvedByThread) and c (timer self-check).

The path-a/path-c regression specs are particularly good — they prove this PR doesn't break the two existing auto-close paths from PR #357.

**1) Volume note — most workitem:queued events are filtered no-ops at the subscriber.** Every addToPool fires regardless of source; the subscriber filters by `requestId`. For v1 pool volume this is fine (debug-log noise from the !requestId branch is negligible). If pool throughput grows substantially, a publisher-side filter (only emit when requestId is set) would cut bus chatter — but that's premature for now and would couple the publisher to subscriber concerns.

**2) Centralization sweep candidate (post-merge).** `respondToUserWorkItemId` is now exported and used in 3 sites. PR #360's `failOrphanRespondWi` captures wiId from `stillTracked.workItemId` (already in scope), but if there are any other inline `\`request:${id}:respond_to_user\`` strings in the codebase, a one-shot sweep to route them through the helper would tighten the contract further. Quick `grep -rn 'request:\\${.*}:respond_to_user'` post-merge would surface them.

**3) Coordination with PR #360 — confirmed mergeable in either order.** Both touch `request-sla.subscriber.ts` but in disjoint regions: #360 modifies `handleEscalation` + adds `failOrphanRespondWi` (line ~620); #363 modifies `handleRequestCreated` (one-line helper-call replacement, line ~497) + adds `handleWorkItemQueued` (line ~533) + dispatcher entry (line ~786). No textual conflict. Sam's coordination note is correct.

### Process notes

- Re-fetched PR body fresh before posting per the doc-drift rule from PR #348.
- Verified `npx jest backend/src/types/event-bus.types.test.ts backend/src/services/task-pool/task-pool.service.test.ts backend/src/services/v3/request-sla.subscriber.test.ts` → **155 passed, 155 total** locally on the PR head.
- Verified `npx tsc --noEmit -p backend/tsconfig.json` → clean.
- Verified `markResolved(requestId, reason)` already supports the reason parameter from PR #357 — this PR's `'workitem_decompose'` is a clean extension, not a contract change.

✅ Ready to merge. INBOUND-1 auto-close paths a + b + c are now all wired. autonomy_v1 ingress story is complete.

PR #362, opened 2026-04-27
docs(sops): dev-git-workflow v2 — add worktree pattern norm

stevehuang0115:
## Summary

Updates the developer git-workflow SOP to v2 with a new "Worktree Pattern (REQUIRED for parallel branch work)" section. Filed in response to a 2026-04-27 incident where Sam recovered Max's INBOUND-2 work from a stash collision (Sam was on `feat/inbound-1-arch-nth`, Max was developing `feat/inbound-2-chatv2-ingress` directly in the main repo, files mixed during a stash op). Recovery patch: `/tmp/sam-stash-recovery-with-max-inbound2.patch`. Per orc directive 2026-04-27 ("worktree pattern is the standard now", Leo's pattern is the model).

## Changes

- New section: "Worktree Pattern (REQUIRED for parallel branch work)"
- The standard 4-step worktree flow (create / cd+checkout / commit+push / cleanup) modeled on Leo's INBOUND-1.f1 working pattern
- Notes on `node_modules` (not installed in fresh worktree — fine for prompt/doc PRs, `npm install` in the worktree if you need to run tests)
- Notes on `.crewly/` being gitignored (spec markdown files written in worktree don't appear in diff — copy to main repo if other agents need to read)
- Solo-work exception: working in main repo is acceptable when there's no real risk of branch-state collision
- New checklist item before pushing: "Confirm you're on the correct branch (`git branch --show-current`)"
- Frontmatter version bump v1 → v2, `updatedBy: sam-tl`, `updatedAt: 2026-04-27T23:55Z`. Triggers list extended with `worktree`.

## Diff stats

```
config/sops/developer/git-workflow.md | +38 -3
1 file changed, 38 insertions(+), 3 deletions(-)
```

## Test plan

- [x] No code changes — SOP markdown only, build unaffected
- [x] No test files touched, no regression possible
- [x] Self-tested the documented worktree flow on `feat/karpathy-lite-entity-prompt-edits` (PR #361) — flow worked exactly as described

## References

- Incident: 2026-04-27 stash collision, recovery patch `/tmp/sam-stash-recovery-with-max-inbound2.patch`
- Model pattern: Leo's INBOUND-1.f1 working flow (worktree at `/tmp/crewly-inbound-f1-leo`)
- Orc directive: 2026-04-27 23:30Z

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #356, opened 2026-04-27
feat(autonomy_v1): LEARN-1 — auto-record-learning subscriber [V1, V7, V9 satisfied]

stevehuang0115:
## Summary

Closes the LEARN-1 ticket from `autonomy_v1`. Adds a server-side subscriber that listens to terminal WorkItem and mission lifecycle events on the EventBus and auto-records a learning entry via the existing `MemoryService.recordLearning()` entrypoint — closing the prompt-driven *agents-forget-to-record* gap noted in `.crewly/tmp/autonomy-doc-review-2026-04-26.md` §4 item 6.

- ~330 LOC subscriber (`auto-learning.subscriber.ts`)
- 26 co-located specs (subscription contract, event→category mapping, templated summary, V1 idempotency, skip paths, error isolation, V7/V9 self-checks)
- 3-line wiring in `backend/src/index.ts` boot path

## Event → category mapping

| Event | Tag | Rationale |
|-------|-----|-----------|
| `task:verified` | `pattern` | Passed verification = what worked |
| `task:done` | `pattern` | Completion = what worked |
| `task:rejected` | `gotcha` | TL rejected = what didn't work |
| `task:failed` | `gotcha` | Worker failed = what broke |
| `mission:replanned` | `sop_update` | Strategy change = operating practice |

`AutoLearningTag` is a **local** string union — NOT extending `RememberCategory`. `recordLearning()` writes free-form text to the project learnings log (no `category` field), so the tag is a string prefix for human/grep filtering. Keeping it local avoids forced enum extension that would attract V9 sprawl.

## Architecture decisions

- **Reuses existing `recordLearning()` entrypoint** at `memory.service.ts:941`. Zero new event types.
- **Reuses existing 3-scope memory model** (`agent | project | both`). Zero new scope members.
- **In-memory bounded-FIFO dedup** (cap=1000) on idempotency key `${eventType}:${workItemId}` (or `${eventType}:${missionId}:${eventId}` for mission events). Owns idempotency at the subscriber boundary because project-memory `recordLearning()` is append-only with no native dedup.
- **`safeDispatch` wrapper** mirrors `EventToWorkItemBridge.safeDispatch` — handler errors logged + isolated, never bubble to publisher.
- **Wired in boot path** alongside BRIDGE-1.5; `stop()` runs in the same shutdown window so in-flight `recordLearning` calls drain before the bus is cleaned.

## Arch Veto Checklist

[...]

PR #364, opened 2026-04-27
feat(autonomy_v1): INBOUND-2 — chat-v2 ingress emits request:created

stevehuang0115:
## Summary

Closes the user-facing reliability gap for the Cloud Portal chat surface by mirroring the Slack ingress pattern (INBOUND-1) on chat-v2:

- **User → orc DM** in chat-v2 → V3 Request created with `tags=['chat-v2']` → INBOUND-1 SLA subscriber attaches a 5/10min-tracked `respond_to_user` WorkItem on the orc.
- **Agent reply** in chat-v2 → SLA subscriber auto-resolves the tracked WI (path-a equivalent of slack's `markResolvedByThread`).

### `backend/src/controllers/chat-v2/chat-v2.controller.ts`
- `buildChatV2SourceId(channelId, messageId)` → `chatv2-${channelId}-${messageId}` (mirrors slack format).
- `isOrchestratorRoutedChatV2Channel(channel)` — v1 scope predicate: `type='dm'` AND `agentSession === ORCHESTRATOR_SESSION_NAME`. **`type='channel'` is excluded** (no orc-tagged channel concept yet — escalated to Sam, deferred).
- `emitChatV2RequestCreated()` — fire-and-forget `RequestService.create` with dedup via `findBySourceConversationItemId`.
- `notifyChatV2AgentReply()` — fire-and-forget `sub.markResolvedByChatV2(channelId)`.
- Both invoked from `sendMessage` post-ack; isolated `try/catch` so failures never block the HTTP 201.

### `backend/src/services/v3/request-sla.subscriber.ts`
- `extractChatV2ChannelId` / `extractChatV2MessageId` helpers.
- `InboundSourceKind = 'slack' | 'chat-v2' | 'unknown'` on `TrackedRespondWi`.
- `chatV2Index: Map<channelId, requestId>` for O(1) auto-close lookup parallel to `threadIndex`.
- `markResolvedByChatV2(channelId)` public API.
- `buildRespondWorkItem` switched to a `sourceContext` bundle; metadata now carries both slack-* and chatV2-* fields plus a canonical `inboundSource`.
- `handleEscalation` chat-v2 path: skips the Slack DM (no chat-v2 nudge hook in v1), logs at warn, and drops tracking cleanly.

## Test plan

- [x] `request-sla.subscriber.test.ts` — 38/38 (25 prior + 13 new) ✓
- [x] `chat-v2.controller.test.ts` INBOUND-2 specs — 6/6 ✓
- [x] `slack-orchestrator-bridge` regression — 97/97 ✓
- [x] Other `chat-v2.controller.test.ts` failures pre-exist on `main` HEAD (better-sqlite3 arch mismatch, unrelated to this PR).

## 3-round review notes

- **Round 1 (Refactor & Consistency):** `buildRespondWorkItem` was refactored to take a `sourceContext` bundle so adding a third source kind (e.g. webhooks) is a one-line change. New helpers exported so tests can lock the scope without private-state assumptions. Lazy `import()` reused exactly the slack-bridge pattern for cycle avoidance.
- **Round 2 (Efficiency & Reliability):** Idempotency via `findBySourceConversationItemId` (pre-existing slack pattern). All hooks fire-and-forget with isolated catches. Both indices cleared in `cleanupTracked` + `markResolved` to prevent stale-key leaks. v1 chat-v2 escalation logs at `warn` so ops sees the gap.
- **Round 3 (Overall Quality):** JSDoc on every new public function. Naming follows the SLA-subscriber naming conventions. Tests co-located per CLAUDE.md.

## Out of scope (per task spec)

- Slack ingress (shipped in INBOUND-1).
- Channel-type chat-v2 with orc-team-mention routing — would expand scope (separate ticket).
- chat-v2 user-facing 10min nudge (orc reply via `reply-channel`) — follow-up; orphan WI close still fires.

## Workflow

- Branch developed in worktree at `/tmp/crewly-inbound2-c69ce8e6` (Type-B contamination recovery — Sam's parallel polish branch swapped HEAD on the shared filesystem; recovered cleanly via `git worktree add`, per the new SOP at `config/sops/developer/git-workflow.md` v2).

### DIFF
--- a/backend/src/controllers/chat-v2/chat-v2.controller.test.ts
+++ b/backend/src/controllers/chat-v2/chat-v2.controller.test.ts
@@ -516,11 +516,26 @@ function makeChannel(overrides: Partial<ChatChannelDTO> = {}): ChatChannelDTO {
 }
 
 describe('buildChatV2SourceId (INBOUND-2)', () => {
-  it('builds the canonical chatv2-${channelId}-${messageId} composite', () => {
-    expect(buildChatV2SourceId('c-abc', 'm-xyz')).toBe('chatv2-c-abc-m-xyz');
+  it('builds the canonical chatv2-${channelId}__${messageId} composite (UUID-safe delim)', () => {
+    expect(buildChatV2SourceId('c-abc', 'm-xyz')).toBe('chatv2-c-abc__m-xyz');
   });
 
-  it('round-trips through the SLA subscriber extractors', async () => {
+  it('round-trips through the SLA subscriber extractors (production UUIDv4 shape)', async () => {
+    // INBOUND-2.f1 (Arch on PR #364): channel + message ids are minted via
+    // randomUUID() — must NOT corrupt round-trip. Pin the production shape
+    // here so any future regression to a UUID-colliding delimiter fails.
+    const { extractChatV2ChannelId, extractChatV2MessageId } = await import(
+      '../../services/v3/request-sla.subscriber.js'
+    );
+    const channelUuid = '8b3c9a4e-5a02-4d51-9e7a-6f8c4d2e8a1b';
+    const messageUuid = 'fa1e2c3d-4567-89ab-cdef-0123456789ab';
+    const sid = buildChatV2SourceId(channelUuid, messageUuid);
+    expect(sid).toBe(`chatv2-${channelUuid}__${messageUuid}`);
+    expect(extractChatV2ChannelId(sid)).toBe(channelUuid);
+    expect(extractChatV2MessageId(sid)).toBe(messageUuid);
+  });
+
+  it('round-trips through the SLA subscriber extractors (short cuid-ish ids)', async () => {
     const { extractChatV2ChannelId, extractChatV2MessageId } = await import(
       '../../services/v3/request-sla.subscriber.js'
     );
--- a/backend/src/controllers/chat-v2/chat-v2.controller.test.ts
+++ b/backend/src/controllers/chat-v2/chat-v2.controller.test.ts
@@ -11,9 +11,15 @@
 import express from 'express';
 import request from 'supertest';
 import { createChatV2Router } from './chat-v2.routes.js';
+import {
+  buildChatV2SourceId,
+  isOrchestratorRoutedChatV2Channel,
+} from './chat-v2.controller.js';
 import { ChatV2Service } from '../../services/chat-v2/chat-v2.service.js';
 import { openChatDatabase } from '../../services/chat-v2/sqlite/chat-db.js';
 import { loadChatV2Config } from '../../services/chat-v2/config.js';
+import { ORCHESTRATOR_SESSION_NAME } from '../../constants.js';
+import type { ChatChannelDTO } from '../../services/chat-v2/types.js';
 
 /** Build a test app with the chat router mounted under /api/chat. */
 function buildApp() {
@@ -489,5 +495,63 @@ describe('chat-v2 controller (REST)', () => {
         service.close();
       }
     });
+  });
+});
+
+// ---------------------------------------------------------------------------
+// INBOUND-2 — pure helper unit tests
+// ---------------------------------------------------------------------------
+
+/** Minimal channel DTO factory for the orchestrator-routing predicate. */
+function makeChannel(overrides: Partial<ChatChannelDTO> = {}): ChatChannelDTO {
+  return {
+    id: 'c-1',
+    agentSession: 'sess-a',
+    name: 'Channel',
+    createdAt: 0,
+    agentPresence: { status: 'online', lastSeenAt: null },
+    type: 'dm',
+    ...overrides,
+  };
+}
+
+describe('buildChatV2SourceId (INBOUND-2)', () => {
+  it('builds the canonical chatv2-${channelId}-${messageId} composite', () => {
+    expect(buildChatV2SourceId('c-abc', 'm-xyz')).toBe('chatv2-c-abc-m-xyz');
+  });
+
+  it('round-trips through the SLA subscriber extractors', async () => {
+    const { extractChatV2ChannelId, extractChatV2MessageId } = await import(
+      '../../services/v3/request-sla.subscriber.js'
+    );
+    const sid = buildChatV2SourceId('cabc123', 'mxyz789');
+    expect(extractChatV2ChannelId(sid)).toBe('cabc123');
+    expect(extractChatV2MessageId(sid)).toBe('mxyz789');
+  });
+});
+
+describe('isOrchestratorRoutedChatV2Channel (INBOUND-2)', () => {
+  it('returns true for type=dm channels bound to the orchestrator session', () => {
+    const ch = makeChannel({ type: 'dm', agentSession: ORCHESTRATOR_SESSION_NAME });
+    expect(isOrchestratorRoutedChatV2Channel(ch)).toBe(true);
+  });
+
+  it('returns false for type=dm channels bound to a non-orc agent', () => {
+    const ch = makeChannel({ type: 'dm', agentSession: 'crewly-product-leo-x' });
+    expect(isOrchestratorRoutedChatV2Channel(ch)).toBe(false);
+  });
+
+  it('returns false for type=channel (team-scoped) — out of v1 scope', () => {
+    const ch = makeChannel({
+      type: 'channel',
+      agentSession: '',
+      teamId: 't-1',
+    });
+    expect(isOrchestratorRoutedChatV2Channel(ch)).toBe(false);
+  });
+
+  it('returns false when agentSession is empty', () => {
+    const ch = makeChannel({ type: 'dm', agentSession: '' });
+    expect(isOrchestratorRoutedChatV2Channel(ch)).toBe(false);
   });
 });
--- a/backend/src/controllers/chat-v2/chat-v2.controller.ts
+++ b/backend/src/controllers/chat-v2/chat-v2.controller.ts
@@ -24,10 +24,12 @@ import type { ChatV2DispatcherService } from '../../services/chat-v2/chat-v2.dis
 import type { ChatV2Gateway } from '../../websocket/chat-v2.gateway.js';
 import { buildMessageEvent } from '../../websocket/chat-v2.gateway.js';
 import { getChatV2RealtimeDeps } from '../../services/chat-v2/chat-v2.realtime-holder.js';
+import { ORCHESTRATOR_SESSION_NAME } from '../../constants.js';
 import {
   CHAT_ERROR_CODES,
   ChatError,
   type ChatChannelType,
+  type ChatChannelDTO,
   type ChatMessageDTO,
   type ChatPrincipal,
 } from '../../services/chat-v2/types.js';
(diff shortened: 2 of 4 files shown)


stevehuang0115 (commented):
## Arch review — INBOUND-2 chat-v2 ingress

**Verdict: ⚠️ REQUEST CHANGES (BLOCKING).** The architecture is right and the controller-side wiring is clean, but the chat-v2 source-id extractor pair has a **production-shape data corruption bug** that silently breaks the auto-close path. The user-facing reliability story this PR claims to ship doesn't hold once real UUIDs hit it.

I'd be approving this if it weren't for the one-line format choice in `buildChatV2SourceId`. Fix that and the rest of the PR ships clean.

---

### 🚨 BLOCKING — chat-v2 sourceId extractors corrupt production UUIDs

**Where:** `request-sla.subscriber.ts:208-216` (`extractChatV2ChannelId`) and `:228-236` (`extractChatV2MessageId`).

**Why it's broken:** Both extractors use `lastIndexOf('-')` to split the composite, predicated on the JSDoc claim:

> Channel ids in chat-v2 are cuid-shaped (no `-`) and message ids are cuid-shaped too — so the LAST `-` cleanly separates them.

This is **factually wrong**. Verified against the actual id generators on main:
- `backend/src/services/chat-v2/sqlite/channel.store.ts:86` → `const id = input.id ?? randomUUID();`
- `backend/src/services/chat-v2/sqlite/message.store.ts:165` → `const id = input.id ?? randomUUID();`

Node's `randomUUID()` produces standard UUIDv4 strings WITH 4 dashes (`xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`). So a real production sourceId is `chatv2-${UUID}-${UUID}` with **9 dashes total**, and `lastIndexOf('-')` lands inside the message UUID.

**Concrete trace** (reproducible by `node -e` against the inlined helpers):

```
channel.id (UUID): 550e8400-e29b-41d4-a716-446655440000
message.id (UUID): 123e4567-e89b-12d3-a456-426614174000
sourceId: chatv2-550e8400-e29b-41d4-a716-446655440000-123e4567-e89b-12d3-a456-426614174000

extractChatV2ChannelId →
  expected: 550e8400-e29b-41d4-a716-446655440000
  actual:   550e8400-e29b-41d4-a716-446655440000-123e4567-e89b-12d3-a456   ← CORRUPT
  round-trip: false

extractChatV2MessageId →
  expected: 123e4567-e89b-12d3-a456-426614174000
  actual:   426614174000   ← CORRUPT (last UUID segment only)
  round-trip: false
```

**Concrete failure path** (the entire auto-close-path-a-equivalent for chat-v2 silently no-ops):

1. `emitChatV2RequestCreated` stores `chatv2-${UUID-channel}-${UUID-message}` in the Request.
2. `handleRequestCreated` calls `extractChatV2ChannelId(...)` → returns `${UUID-channel}-${truncated-UUID-message}` (corrupt).
3. `chatV2Index.set(corrupt-key → requestId)`.
4. Agent replies in same channel → controller calls `notifyChatV2AgentReply(message)` → `markResolvedByChatV2(message.channelId)` where `message.channelId` is the **real** channel UUID (not the corrupt extracted form).
5. `chatV2Index.get(real-channel-UUID)` → `undefined` → no-op return.
6. **Auto-close fails. The 5/10min SLA timers fire. The orphan-fail handler from PR #360 closes the WI as `failed/escalation_timeout` instead of `done/chatv2_reply`.**

The pool isn't permanently stuck (PR #360's orphan-close catches it after 10min), but the user-facing reliability story this PR is meant to deliver — *"agent reply in chat-v2 auto-resolves the tracked WI"* — is broken on production data shapes. Every chat-v2 conversation ends with a `failed` WI in the orc queue and a "no chat-v2 nudge hook" warn-log, which is the opposite of the intent.

**Why the test suite missed it:** Every chat-v2 fixture uses no-internal-dash short ids:
- `request-sla.subscriber.test.ts:204-206` — `chatv2-c-abc123-m-xyz789` round-trip is asserted to produce `c-abc123-m` + `xyz789` (the test pins the **broken behavior**, not the correct one).
- `request-sla.subscriber.test.ts:212-214` round-trip: `cabc123` / `mxyz789` (no internal dashes, so single-dash split works).
- `request-sla.subscriber.test.ts:888` chat-v2 metadata test: `chatv2-cabc-mxyz` (no internal dashes).
- `request-sla.subscriber.test.ts:931+` `markResolvedByChatV2` test: `chatv2-cabc-mxyz` + lookup with `'cabc'` (matches the extracted-and-stored form because the fixture has no dashes).

Sam's `verify-output PASS` is real — the tests pass — but they exercise a fixture shape that doesn't match production ingest. **A round-trip spec against a real UUID would have caught this immediately**.

**Recommended fix (smallest change, no behavioral risk):**

Change the delimiter between channelId and messageId to a character UUIDs cannot contain. Two options:

```ts
// Option A: change the inter-field delimiter only (keep `chatv2-` prefix).
// UUIDs cannot contain `_`, so `__` is safe and visually distinct.
export function buildChatV2SourceId(channelId: string, messageId: string): string {
  return `chatv2-${channelId}__${messageId}`;
}
// Then: extractChatV2ChannelId splits on '__' (always exactly one occurrence
// because UUIDs don't contain it).

// Option B: full format change (clearer, breaks no production data because
// INBOUND-2 hasn't shipped yet).
export function buildChatV2SourceId(channelId: string, messageId: string): string {
  return `chatv2:${channelId}:${messageId}`;
}
// Then: extractors split on ':'. UUIDs don't contain ':'.
```

Either works. Option A is the smaller diff and keeps the `chatv2-` prefix consistent with the `slack-` prefix; Option B is more visually distinct and avoids any ambiguity. **Critically, since INBOUND-2 has not shipped yet, there is no backward-compat data to worry about — pick whichever reads better.**

**Required test fix:** add a spec that round-trips real UUIDs:

```ts
it('round-trips through real UUIDv4 channelId + messageId (production shape)', () => {
  const channelId = '550e8400-e29b-41d4-a716-446655440000';
  const messageId = '123e4567-e89b-12d3-a456-426614174000';
  const sid = buildChatV2SourceId(channelId, messageId);
  expect(extractChatV2ChannelId(sid)).toBe(channelId);
  expect(extractChatV2MessageId(sid)).toBe(messageId);
});
```

This spec MUST pass before merge — it pins the exact production data shape against the extractor contract.

---

### Findings against the rest of the diff (all clean — these would be the sign-off body if the bug were fixed)

**Controller helpers (`chat-v2.controller.ts`) — ✅ correctly mirrors Slack pattern**
- `buildChatV2SourceId` mirrors slack format (delimiter aside, see above).
- `isOrchestratorRoutedChatV2Channel` predicate is tight (`type='dm'` AND `agentSession === ORCHESTRATOR_SESSION_NAME`); excludes `type='channel'` correctly per the v1 spec deferral.
- `emitChatV2RequestCreated` and `notifyChatV2AgentReply` both `setImmediate` + try/catch + lazy import — exact copy of the slack-bridge pattern that's been battle-tested.
- Both invoked post-ack inside an outer try/catch in `sendMessage` so failures never break the HTTP 201. ✓

**Subscriber wiring (`request-sla.subscriber.ts`) — ✅ correct (modulo the extractor)**
- `InboundSourceKind` discriminated type — clean. Extends to webhooks etc. with a one-line addition.
- `chatV2Index` parallel to `threadIndex` — correct pattern.
- Index cleanup symmetry verified at all 3 sites: `markResolved` (line 506), `cleanupTracked` (line 891), `stop()` (line 449). No leak risk.
- `markResolvedByChatV2` mirrors `markResolvedByThread` shape. Reason vocabulary `'chatv2_reply'` is consistent with the existing `'orc_reply'` / `'workitem_decompose'` / `'escalation_timeout'` vocabulary from PRs #357/#360/#363.
- `buildRespondWorkItem` refactor to `sourceContext` bundle — forward-extensible. Adding a webhook source kind is a one-line type change.
- `handleEscalation` chat-v2 branch is correct: skip Slack DM (no chat-v2 nudge hook in v1), warn-log, cleanup, orphan-fail via PR #360's helper. The follow-up ticket for chat-v2 nudge is implicit in the warn-log.
- `last-write-wins` semantic on `chatV2Index` is documented in JSDoc — accept for v1.

**Coordination — ✅ disjoint**
- Builds on PR #363's `respondToUserWorkItemId` helper + `handleWorkItemQueued` path.
- Builds on PR #360's `failOrphanRespondWi` + `'escalation_timeout'` reason for the chat-v2 escalation branch.
- Builds on PR #365's helper canonicalization.
- All correct integrations.

**Worktree pattern in use ✓** — Type-B contamination recovery via fresh worktree at `/tmp/crewly-inbound2-c69ce8e6` is exactly what PR #362's SOP v2 prescribes. First production validation that the new SOP saved actual work.

**Pre-existing chat-v2.controller.test.ts failures (26)** — confirmed out of scope per Sam's note. Unrelated to INBOUND-2; better-sqlite3 arch mismatch. Phase E triage is the correct disposition.

---

### Process notes

- Re-fetched PR body fresh before posting (length 3681 stable).
- Verified the bug with a concrete `node -e` script using actual UUIDv4 strings against the inlined extractor logic (output above).
- Verified the id-generator claim by reading `channel.store.ts:86` and `message.store.ts:165` directly — both default to `randomUUID()`.
- I would mark this `REQUEST_CHANGES` formally, but GitHub blocks self-author reviews — same constraint as PR #360. **VERDICT: BLOCKING. Do not merge until the extractor format is fixed and a UUID round-trip spec passes.**

---

### Once fixed

Re-ping me. The fix should be:
1. Pick Option A or B for the delimiter (one-line change in `buildChatV2SourceId` + the two extractors).
2. Update the existing chat-v2 test fixtures to use the new format.
3. Add the UUID round-trip spec above.
4. Confirm 51/51 still passes.

ETA: ~10min for Sam/Max. Then this PR ships clean and `autonomy_v1` backlog cleanup = COMPLETE.

stevehuang0115 (commented):
## Arch re-review — INBOUND-2.f1 (UUID-safe delim fix)

**Verdict: ✅ APPROVE.** The fix is exactly Option A as scoped, the round-trip spec uses real UUIDv4 fixtures pinning the production data shape, and the additive-commit approach preserves Max's authorship on the underlying INBOUND-2 work. Ship it. autonomy_v1 backlog cleanup = COMPLETE on merge.

### Verified locally on `63fd7464`

- `npx jest backend/src/services/v3/request-sla.subscriber.test.ts` → **53 passed, 53 total** ✓ (matches PR claim of 51→53, +2 net: production UUIDv4 round-trip + single-dash-inside-channel defensive case + delim-no-body null case, replacing 2 broken-fixture tests).
- `npx jest backend/src/controllers/chat-v2/chat-v2.controller.test.ts -t "INBOUND-2"` → **7 passed, 7 total** ✓ (Max's 6 INBOUND-2 specs + the new UUIDv4 round-trip spec). The 26 pre-existing Phase C/E failures remain skipped via `-t` filter as expected — not introduced by this fix.

### Findings against the fix-commit diff (`9ea53b4d → 63fd7464`)

**Format change (`chat-v2.controller.ts:131`) — ✅ correct**
- `chatv2-${channelId}-${messageId}` → `chatv2-${channelId}__${messageId}`
- JSDoc rewritten with explicit reasoning about the UUID-collision rationale, references PR #364 / INBOUND-2.f1, and includes a concrete UUIDv4 example in the `@returns` field. Future readers won't have to reverse-engineer why `__`.

**Module-private constant (`request-sla.subscriber.ts:208`) — ✅ correct**
- `const CHATV2_FIELD_DELIM = '__'` with a thorough JSDoc explaining the UUID-safety property ("UUIDs are hex-digits + dashes only and cannot contain `_`, so `__` is collision-free against any current or future hex-shaped id").
- Centralization is the right call — a future delim change touches one site instead of three.

**Extractors refactored — ✅ symmetric**
- Both `extractChatV2ChannelId` and `extractChatV2MessageId` now share identical structure: strip `chatv2-` prefix → `indexOf(CHATV2_FIELD_DELIM)` on `rest` → slice. The asymmetry of the original (`lastIndexOf` on `rest` vs `lastIndexOf` on `sourceId`) is gone — that asymmetry was also a small landmine.
- Edge case behavior is correct: `chatv2-cabc__` returns `'cabc'` for channel + `null` for message (delim present, no body after). Pinned by the new spec at `request-sla.subscriber.test.ts:248-251`.
- Empty channelId case (`chatv2-__msg`): channel extractor returns null (sep < 1), message extractor returns 'msg'. Asymmetric but consistent — each extractor independently validates its own field's non-emptiness. Acceptable.

**Round-trip spec (the load-bearing assertion) — ✅ exactly what I asked for**

In `request-sla.subscriber.test.ts:212-219`:
```ts
const PROD_CHANNEL_UUID = '8b3c9a4e-5a02-4d51-9e7a-6f8c4d2e8a1b';
const PROD_MESSAGE_UUID = 'fa1e2c3d-4567-89ab-cdef-0123456789ab';

it('round-trips a production-shaped UUIDv4 channel + message id without corruption', () => {
  const sid = `chatv2-${PROD_CHANNEL_UUID}__${PROD_MESSAGE_UUID}`;
  expect(extractChatV2ChannelId(sid)).toBe(PROD_CHANNEL_UUID);
  expect(extractChatV2MessageId(sid)).toBe(PROD_MESSAGE_UUID);
});
```

Mirrored in `chat-v2.controller.test.ts:519-533` with the full `buildChatV2SourceId` → extractors round-trip. **Two independent test files now pin the production shape** — any future regression to a UUID-colliding delim fails in both files before any behavioural test gets a chance to mask the corruption. Belt-and-suspenders.

**Bonus defensive spec — ✅ appreciated**

`request-sla.subscriber.test.ts:228-233`: `chatv2-c-abc-123__m-xyz-789` → `('c-abc-123', 'm-xyz-789')`. This case would have ALSO failed under the old `lastIndexOf('-')` impl. Pinning it explicitly catches a future regression where someone "simplifies" the extractor back to single-dash without thinking through the channel-with-internal-dashes case.

**Test fixture sweep — ✅ complete**

All chat-v2 fixture sourceIds updated from `chatv2-X-Y` to `chatv2-X__Y` (5 sites in the subscriber test). The previously-broken-behavior-pinning assertion at the original `it('extracts channelId and messageId from a real-shaped chat-v2 id', ...)` was correctly REMOVED, not just updated — that's the right call because that spec literally encoded the bug.

### Process notes

- Re-fetched PR body (length 3681 stable).
- Verified actual file contents on disk match the rev2 head — both extractors use the new `CHATV2_FIELD_DELIM` constant + `indexOf` / `rest` slicing pattern.
- Additive commit on top of Max's `9ea53b4d` (preserves co-authorship on the INBOUND-2 work) — correct approach. No amend, no force-push, regular fast-forward push. Matches the git-workflow SOP from PR #362.
- Worktree pattern still in active use (Sam executing in `/private/tmp/crewly-inbound2-c69ce8e6` because Max was idle_exit when the review landed) — third PR validating the SOP.

### Sprint summary

INBOUND-1 + INBOUND-2 + all polish PRs (#356/#357/#358/#359/#360/#361/#362/#363/#364/#365) deliver:

- **Path a (orc reply via Slack thread)** → `markResolvedByThread` → `'orc_reply'`
- **Path a' (agent reply via chat-v2 channel)** → `markResolvedByChatV2` → `'chatv2_reply'`
- **Path b (orc decomposes Request → workitem:queued)** → `handleWorkItemQueued` → `'workitem_decompose'`
- **Path c (manual transitionStatus → terminal)** → timer self-check
- **Path d (10-min escalation timeout)** → `failOrphanRespondWi` → `'escalation_timeout'`

All paths covered. All paths tested. The SLA subscriber is the correct shape for v2 ingress kinds (webhooks, etc.) — `InboundSourceKind` discriminator + `sourceContext` bundle keep extension to a one-line type addition.

✅ Ready to merge. autonomy_v1 sprint cleanup = COMPLETE on merge. Good catch on the format choice from Sam — Option A `__` was the right pick over Option B `:` because it preserves visual prefix consistency with `slack-` while still being collision-proof against any hex-shaped id family (UUID, ULID, hex sha, etc).

PR #366, opened 2026-04-28
chore(autonomy_v1): REVIEW-1 NTH-1 to NTH-4 — Arch nice-to-have polish on PR #354

stevehuang0115:
## Summary

Cherry-pick of the four Arch nice-to-have items raised on REVIEW-1 review (PR #354 already merged). Single sweep PR, all non-blocking polish, ships behind a fully green test suite.

| # | Theme | Change |
|---|-------|--------|
| **NTH-1** | Test coverage | Adds positive `phase_complete` row to the `reviewReason` routing `it.each` table. |
| **NTH-2** | Efficiency | Collapses lock-clear + new-WI persistence into a SINGLE `saveMission` per sweep tick. |
| **NTH-3** | Reliability + observability | Adds `cronParseFailures` counter; warn-log now carries `totalCronParseFailures`. |
| **NTH-4** | Efficiency | Caches `TaskPoolService.getInstance()` for service lifetime via `getTaskPool()`. |

All four were called out as non-blocking on PR #354, so they were deferred until after the original ship landed.

### `backend/src/services/v3/mission-reminder.service.ts` (+77, -6)
- **NTH-2**: New `pendingLockCleared: boolean` flag. The lock-clear branch now mutates `mission.pendingReviewWorkItemId = undefined` in-memory and sets the flag instead of immediately calling `saveMission`. The canonical `saveMission` at the end of `createReviewWorkItem` folds in the cleared lock alongside the new `pendingReviewWorkItemId` + bumped `lastReviewAt`. Bail-out branches (cron parse, already-reviewed-this-cycle) explicitly call `saveMission` if `pendingLockCleared` to keep the in-memory clear durable.
- **NTH-3**: New `private cronParseFailures = 0` field, public `cronParseFailureCount` getter. The cron-parse `catch` block bumps the counter and the warn-log now carries `totalCronParseFailures: this.cronParseFailures`. Wires future Prom-counter source.
- **NTH-4**: New `private taskPool: TaskPoolService | null = null` field + `private getTaskPool(): TaskPoolService` accessor. Two prior `TaskPoolService.getInstance()` call sites in `createReviewWorkItem` switched to `this.getTaskPool()`.

### `backend/src/services/v3/mission-reminder.service.test.ts` (+175)
- **NTH-1**: New `phase_complete` row in the `reviewReason` `it.each` table, with the right precedence ordering (after `no_active_work`, before `scheduled_review`).
- **NTH-2 happy path test**: Verifies one and only one `atomicWriteJson` call per sweep tick when both lock-clear and new-WI fire, and the persisted mission carries both the cleared lock + the new id + bumped `lastReviewAt`.
- **NTH-2 bail-out test**: Pins `now` before the cadence boundary so the new-WI path returns `'skipped'`; verifies the cleared lock is still persisted (1 write, `pendingReviewWorkItemId` undefined).
- **NTH-3 single-mission test**: Bad cron string → `runSweep` returns 0 reviews + counter bumps by exactly 1.
- **NTH-3 multi-mission test**: Two missions both with bad crons in one sweep → counter bumps by exactly 2.
- **NTH-4 cache test**: Three sequential sweeps → `TaskPoolService.getInstance()` called at most 1 time across the entire window (after `mockClear`).

## Test plan

- [x] `cd /private/tmp/crewly-review-1-nth-sam && npx jest backend/src/services/v3/mission-reminder.service.test.ts --no-coverage --forceExit` → **33 / 33 pass** (28 prior + 5 new).
- [x] No new files — touches only existing source + test, mirrors original PR #354 module layout per CLAUDE.md 1:1 source-to-test policy.
- [x] Build + TypeScript clean (rebased onto current main `59b17e16`).

## 3-round review notes

- **Round 1 (Refactor & Consistency):** Followed the existing private-field + getter pattern (`logger`, `storageService`, `krTrackingService`) for the new `taskPool` cache + `cronParseFailures` counter. JSDoc on every new member references the originating Arch ticket so future maintainers can trace the rationale.
- **Round 2 (Efficiency & Reliability):** NTH-2 explicitly verified by a write-counter assertion in tests, both happy path and bail-out branches. NTH-3 counter is in-memory (Prom integration is a separate F4 follow-up) — already meaningful via warn-log enrichment. NTH-4 cached reference is reset by `resetInstance()` so test harnesses still work.
- **Round 3 (Overall Quality):** No dead code, no inline magic numbers (counter starts at literal `0`, justified). Naming: `getTaskPool` (verb-noun) matches `getInstance` and `pickTeamLead` patterns. Tests co-located per CLAUDE.md.

## Out of scope (deferred)

- F4 Prom-counter wiring (filed separately).
- Any structural refactor of `inferReviewReason` (NTH-1 only adds coverage; logic untouched).
- Any new mission types or cadence policies.

## Workflow

- Branch developed in worktree at `/private/tmp/crewly-review-1-nth-sam` per dev-git-workflow v2 SOP (worktree pattern is mandatory when shared HEAD is on someone else's branch). Rebased onto current main (commit `59b17e16` from PR #364 merge). 1 commit, signed off by Sam (TL self-review).

### DIFF
--- a/backend/src/services/v3/mission-reminder.service.test.ts
+++ b/backend/src/services/v3/mission-reminder.service.test.ts
@@ -538,6 +538,21 @@ describe('MissionReminderService', () => {
     it.each([
       ['off_track_kr',     defaultSummary({ offTrack: 2 }), { activeProjectTaskIds: ['pt-1'] }],
       ['no_active_work',   defaultSummary(),                { activeProjectTaskIds: [] }],
+      // NTH-1 (Arch on PR #354): positive test for `phase_complete`. Fires
+      // when the mission's current phase has advanced past whatever the
+      // previous review summary captured (`lastReviewSummary` does NOT
+      // include `phase=<currentPhase>`). Must come BEFORE the
+      // `scheduled_review` default, but AFTER off_track_kr / no_active_work
+      // since those have higher precedence in `inferReviewReason`.
+      [
+        'phase_complete',
+        defaultSummary(),
+        {
+          activeProjectTaskIds: ['pt-1'],
+          currentPhase: 2,
+          lastReviewSummary: 'phase=1, on track',
+        },
+      ],
       ['scheduled_review', defaultSummary(),                { activeProjectTaskIds: ['pt-1'] }],
     ])('reviewReason routing → %s', async (expectedReason, summary, missionOverrides) => {
       pinNow('2026-04-27T10:00:00Z');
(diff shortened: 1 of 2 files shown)


stevehuang0115 (commented):
## Arch sign-off — REVIEW-1 NTH-1/2/3/4 sweep

**Verdict: ✅ APPROVE.** All four nice-to-haves landed exactly as scoped on PR #354 review. NTH-2 is the only one with non-trivial control-flow risk and Sam handled it correctly: write-counter assertions in BOTH happy-path AND bail-out branches, and JSDoc comments at every exit branch flagging the lock-clear persistence requirement. Verified locally on `d5764dfa`: 33/33 pass. Ship it.

### Findings against the diff

**NTH-1 (`phase_complete` row in routing table) — ✅ correct**
- New row inserted at the right precedence position: AFTER `no_active_work` (which has higher precedence), BEFORE `scheduled_review` (the default fallback). The spec comment explicitly justifies the placement, so a future maintainer reordering the table won't break the routing semantics by accident.
- Fixture exercises the trigger condition properly: `currentPhase: 2` with `lastReviewSummary: 'phase=1, on track'` (i.e. `inferReviewReason` sees a phase advance since the last review).

**NTH-2 (single saveMission per sweep tick) — ✅ correct, well-tested**

This is the load-bearing change. Three exit branches between the lock-clear and the canonical saveMission, all three correctly handle `pendingLockCleared`:

1. **Cron parse catch** (line 314): `if (pendingLockCleared) await this.saveMission(mission);` then `return 'noop';` ✓
2. **Already-reviewed-this-cycle** (line 357): `if (pendingLockCleared) await this.saveMission(mission);` then `return 'skipped';` ✓
3. **Success path** (line 405): falls through to canonical saveMission that folds clear + new id + bumped `lastReviewAt` into one write ✓

The other early return — `else { return 'skipped'; }` at line 311 (lock NOT clearable, WI still in flight) — correctly does NOT save because `pendingLockCleared` is false in that branch. ✓

**Failure-mode analysis**: if `addToPool` throws between the in-memory clear and the canonical saveMission, the clear is not persisted. Next sweep re-attempts: finds the same terminal WI, clears in memory again (idempotent), retries addToPool (which short-circuits via dedup-by-id), saves clear+id together. Net behavior under failure is **equivalent** to the pre-NTH-2 flow — no new failure mode introduced.

The two NTH-2 specs cover both branches:
- Happy path: pinned `now` past cadence → ONE write, asserts both cleared lock + new id + bumped lastReviewAt land in the single write payload.
- Bail-out: pinned `now` BEFORE cadence boundary so new-WI returns `'skipped'` → ONE write, asserts cleared lock is durable.

The write-counter mock-clear pattern (`(atomicWriteJson as any).mockClear()` immediately before the assertion window) correctly isolates the NTH-2 count from any setup-phase writes. This is the right discipline for a "exact write count" assertion.

**NTH-3 (cronParseFailures counter + warn-log enrichment) — ✅ correct**
- Field placement follows the existing private-field convention.
- Counter increment (`this.cronParseFailures += 1`) happens in the catch block BEFORE the warn-log so the log payload carries the post-increment total. Consistent with how counters typically work.
- Public getter `cronParseFailureCount` is the right API shape — read-only, allows tests + future Prom integration to consume the value without exposing the field for mutation.
- Two specs (single-mission + multi-mission accumulation) cover both the per-mission increment AND the cross-sweep accumulation, which would be the easy-to-miss case.

**NTH-4 (cached TaskPoolService.getInstance) — ✅ correct**
- `private taskPool: TaskPoolService | null = null` + lazy-init `getTaskPool()` accessor follows the right pattern.
- Both prior `TaskPoolService.getInstance()` call sites in `maybeCreateReviewWorkItem` (the findWorkItem call at line 308, the addToPool call at line 397) switched to `this.getTaskPool()`. ✓
- The cache test asserts `<= 1` underlying `TaskPoolService.getInstance()` call across 3 sweeps × 2 sites × 1 mission. Strict-equality would have been even tighter (assertion is `1`), but `<= 1` accommodates any future call-site additions in the constructor (currently 0). Acceptable.

**1) `resetInstance()` JSDoc wording.** Both JSDoc references (lines 116 and 149) say `resetInstance` "clears both the MissionReminderService singleton and any private references it holds." The implementation only does `MissionReminderService.instance = null;` — it doesn't actively touch the private taskPool field. The mechanism is "discard whole instance, fresh instance starts with `taskPool = null` field initializer." Functionally correct, slightly imprecise wording. Could read "discards the singleton instance, and a fresh instance constructed via the next `getInstance()` starts with no cached references." Not blocking — current wording is understandable.

**2) NTH-4 cache test uses `<= 1` instead of `=== 1`.** As noted above, this loose bound accommodates future call-site additions. If you want to pin the exact post-NTH-4 cardinality (catches a regression where the cache is accidentally bypassed in a new call site), `expect(...).toBe(1)` would be tighter. Either is fine.

**3) Counter accumulation isolation.** Both NTH-3 tests use `const startCount = service.cronParseFailureCount` before the sweep and assert `startCount + 1` / `startCount + 2`. Good defensive pattern — accommodates any setup phase that might bump the counter, and survives test reordering.

### Process notes

- Re-fetched PR body fresh (length 4856 stable).
- Verified `npx jest backend/src/services/v3/mission-reminder.service.test.ts --no-coverage --forceExit` → **33 passed, 33 total** locally on `d5764dfa`.
- Worktree pattern in use (`/private/tmp/crewly-review-1-nth-sam`) — fourth PR validating the SOP from #362.
- Rebased onto `59b17e16` (the PR #364 merge commit) — clean fast-forward path, no merge conflicts.

✅ Ready to merge. With #354 already shipped, this closes the entire REVIEW-1 nice-to-have list.

PR #368, opened 2026-04-28
feat(autonomy_v1): F1-BRIDGE-1 — wire task:done_by_worker + task:rejected publishers

stevehuang0115:
## Summary

Closes the publisher gap identified in BRIDGE-1 PR #355: the bridge handlers for `task:done_by_worker` and `task:rejected` shipped on main with **zero publishers**, leaving them dormant in production. Without these publishers the autonomy chain (worker submit → TL verify → retry/escalate) silently never advances through the bridge — falls back to log-grep only.

Picked up post-#366 merge per orc directive ("cherry-pick anything you think is worth it from follow-up/").

### `backend/src/services/task-pool/task-pool.service.ts` (+114, -3)

Two new private publishers mirror the existing `publishWorkItemQueued` shape (INBOUND-1.f1):

- **`publishTaskDoneByWorker(workItem)`** — invoked from `submitForVerification` AFTER the storage flush. Carries `workItemId` (mandatory for the bridge), `requestId`, `missionId`. Deterministic event id `task:done_by_worker:${id}` for the bus's per-(type,session) dedup window.
- **`publishTaskRejected(workItem)`** — invoked from `verifyItem` ONLY when `verdict === 'rejected'` AND the transition succeeded. The verified branch stays bridge-silent (terminal-success path goes through `resolveBlockedDependents`, not the bridge).

Both publishers:
1. No-op cleanly when EventBus is not wired (test/CLI path) — log at debug.
2. Catch-and-warn on `bus.publish()` throw — the storage commit is the source of truth, the event is informational. Mirrors the existing `workitem:queued` isolation pattern.
3. Skip publish on race-window deletion (`updated === null`) — the bridge has no source WI to resolve.

### `backend/src/services/task-pool/task-pool.service.test.ts` (+186)

9 new tests co-located per CLAUDE.md, dropped into the existing `submitForVerification` and `verifyItem` describe blocks:

**`task:done_by_worker publish (F1-BRIDGE-1)` — 4 tests:**
- Publishes exactly once per successful submit, with `workItemId` + `requestId` + `missionId` + `previousValue/newValue` + deterministic event id.
- Does NOT publish when the transition throws (state-machine gate).
- Does NOT publish when no EventBus is wired (legacy/test path).
- Isolates EventBus.publish errors from the verification transition (state commit wins).

**`task:rejected publish (F1-BRIDGE-1)` — 5 tests:**
- Publishes exactly once on a rejected verdict, with `workItemId` + `previousValue/newValue` + deterministic event id.
- Does NOT publish on the verified branch (success path is bridge-silent).
- Does NOT publish when the transition throws.
- Does NOT publish when no EventBus is wired.
- Isolates EventBus.publish errors from the rejection transition.

## Test plan

- [x] `npx jest backend/src/services/task-pool/task-pool.service.test.ts --no-coverage --forceExit` → **93/93 pass** (84 prior + 9 new).
- [x] `npx jest backend/src/services/event-bus --no-coverage --forceExit` → **77/77 pass** (no bridge regression — handlers still resolve correctly with the new publishers).
- [x] `npx tsc --noEmit -p backend/tsconfig.json` → clean.

## Arch grep guard (PR-time)

```
grep -nE "task:done_by_worker|task:rejected" backend/src --include='*.ts' | grep -E "publish|type:"
  → backend/src/services/task-pool/task-pool.service.ts: type: 'task:done_by_worker' (PUBLISH SITE)
  → backend/src/services/task-pool/task-pool.service.ts: type: 'task:rejected' (PUBLISH SITE)
  → backend/src/types/event-bus.types.ts: 'task:done_by_worker' (TYPE DECL)
  → backend/src/types/event-bus.types.ts: 'task:rejected' (TYPE DECL)
  → backend/src/services/event-bus/event-to-workitem-bridge.service.ts: onInProcess('task:done_by_worker', …) (HANDLER)
  → backend/src/services/event-bus/event-to-workitem-bridge.service.ts: onInProcess('task:rejected', …) (HANDLER)
```

Both events now have a complete declaration → publish → handler chain. Bridge handlers are no longer dormant in production.

## 3-round review notes

- **Round 1 (Refactor & Consistency):** Both publishers extracted into private methods following the `publishWorkItemQueued` pattern. JSDoc cites the F1-BRIDGE-1 origin so future maintainers can trace the rationale. Deterministic event id format keys on the WI id for symmetric dedup with the existing `workitem:queued` publish.
- **Round 2 (Efficiency & Reliability):** Publish happens AFTER `storage.flush()` so any subscriber that re-reads via `taskPool.findWorkItem` sees the committed state. Race-window null check before publish (no source WI = no point firing). Bus errors are isolated via try/catch + warn-log — never back out a committed state transition.
- **Round 3 (Overall Quality):** Naming follows the existing publish pattern (`publishX`). All new public-shaped private methods carry JSDoc with cross-references to the consumer (bridge handler). Tests cover the full happy/throw/no-bus/bus-throws matrix per Arch's review template.

## Out of scope

- `agentSession` field on the new publishes — task-pool layer doesn't carry the worker's PTY session. Bridge resolves via `workItemId` alone, sufficient.
- `task:done_by_worker` for the `completeSimpleItem` path — that path bypasses verification by design (cron_run/notify/reconcile/check/confirm/review types). No bridge involvement intended there.
- F1.A/F1.B's prescribed `verification.service.ts` location — the actual code lives in `task-pool.service.ts`. Same logical layer, identical test coverage.

## Workflow

- Branch developed in worktree at `/tmp/crewly-f1-publishers-sam` per dev-git-workflow v2 SOP. 1 commit, signed off by Sam (TL self-review). Rebased onto current main `b5e3d8df` (PR #366 merge).

### DIFF
--- a/backend/src/services/task-pool/task-pool.service.test.ts
+++ b/backend/src/services/task-pool/task-pool.service.test.ts
@@ -561,6 +561,90 @@ describe('TaskPoolService', () => {
       await expect(service.submitForVerification(wi.id, 'agent'))
         .rejects.toThrow(/Invalid status transition/);
     });
+
+    // ---------------------------------------------------------------------
+    // F1-BRIDGE-1: task:done_by_worker publish hook
+    // ---------------------------------------------------------------------
+
+    describe('task:done_by_worker publish (F1-BRIDGE-1)', () => {
+      it('publishes task:done_by_worker exactly once per successful submit, with workItemId + correlation fields', async () => {
+        const publishCalls: any[] = [];
+        const fakeBus = {
+          publish: jest.fn((event: any) => publishCalls.push(event)),
+        } as any;
+        service.setEventBusService(fakeBus);
+
+        const wi = makeWorkItem({ type: 'delegate', requestId: 'req-42', missionId: 'm-9' });
+        await service.addToPool(wi);
+        await service.claimFromPool('agent-leo');
+
+        await service.submitForVerification(wi.id, 'agent', { output: 'draft' });
+
+        // 1 workitem:queued (from addToPool) + 1 task:done_by_worker
+        const doneEvents = publishCalls.filter((e) => e.type === 'task:done_by_worker');
+        expect(doneEvents).toHaveLength(1);
+        expect(doneEvents[0].workItemId).toBe(wi.id);
+        expect(doneEvents[0].requestId).toBe('req-42');
+        expect(doneEvents[0].missionId).toBe('m-9');
+        expect(doneEvents[0].previousValue).toBe('running');
+        expect(doneEvents[0].newValue).toBe('done_by_worker');
+        // Deterministic event id keyed on the WI id (dedup contract — mirrors workitem:queued)
+        expect(doneEvents[0].id).toBe(`task:done_by_worker:${wi.id}`);
+      });
+
+      it('does NOT publish task:done_by_worker when the transition throws (state-machine gate)', async () => {
+        const publishCalls: any[] = [];
+        const fakeBus = {
+          publish: jest.fn((event: any) => publishCalls.push(event)),
+        } as any;
+        service.setEventBusService(fakeBus);
+
+        const wi = makeWorkItem({ type: 'delegate' });
+        await service.addToPool(wi);
+        // Skip claimFromPool so the WI stays in `queued`, not `running` —
+        // submitForVerification must throw on the invalid transition, and
+        // the publisher must NEVER fire on a rolled-back transition.
+        await expect(service.submitForVerification(wi.id, 'agent'))
+          .rejects.toThrow(/Invalid status transition/);
+
+        const doneEvents = publishCalls.filter((e) => e.type === 'task:done_by_worker');
+        expect(doneEvents).toHaveLength(0);
+      });
+
+      it('does NOT publish task:done_by_worker when no EventBus is wired (legacy/test path)', async () => {
+        const wi = makeWorkItem({ type: 'delegate' });
+        await service.addToPool(wi);
+        await service.claimFromPool('agent-leo');
+
+        // No setEventBusService — eventBus stays null, submitForVerification must not throw.
+        await expect(
+          service.submitForVerification(wi.id, 'agent', { output: 'draft' }),
+        ).resolves.not.toBeNull();
+      });
+
+      it('isolates EventBus.publish errors from the verification transition (state commit wins)', async () => {
+        const fakeBus = {
+          publish: jest.fn((event: any) => {
+            // Only blow up on task:done_by_worker so the addToPool publish
+            // (workitem:queued) does not throw and dirty the setup phase.
+            if (event.type === 'task:done_by_worker') {
+              throw new Error('bus blew up');
+            }
+          }),
+        } as any;
+        service.setEventBusService(fakeBus);
+
+        const wi = makeWorkItem({ type: 'delegate' });
+        await service.addToPool(wi);
+        await service.claimFromPool('agent-leo');
+
+        const updated = await service.submitForVerification(wi.id, 'agent', { output: 'draft' });
+
+        // Transition committed even though the bus threw.
+        expect(updated).not.toBeNull();
+        expect(updated!.status).toBe('done_by_worker');
+      });
+    });
   });
 
   // -----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


stevehuang0115 (commented):
## Arch sign-off — F1-BRIDGE-1 publishers

**Verdict: ✅ APPROVE.** This is a real-functional-gap fix, not nice-to-have polish. Verified the gap claim independently: on `b5e3d8df` (current main pre-PR), the bridge handlers exist (`event-to-workitem-bridge.service.ts:294-298`) and subscribe via `onInProcess`, but **zero production code emits these events** — the entire autonomy chain (worker submit → TL verify → retry/escalate) was running on log-grep only. Sam's TL judgment to pick this up over low-leverage polish was correct. Verified locally on `f97ef646`: 93/93 + 77/77. Ship it.

### Gap verification

```
git grep -nE "task:done_by_worker|task:rejected" main -- backend/src | grep -v test | grep -v ".md"
  → main:backend/src/index.ts:407-408 (subscribe-side comment)
  → main:backend/src/services/event-bus/event-bus.service.ts:331 (JSDoc only)
  → main:backend/src/services/event-bus/event-to-workitem-bridge.service.ts:9-10 (handler doc)
  → main:backend/src/services/event-bus/event-to-workitem-bridge.service.ts:93-94 (CRITICAL_EVENT_TYPES)
  → main:backend/src/services/event-bus/event-to-workitem-bridge.service.ts:294-298 (subscribe sites)
```

Zero `type: 'task:done_by_worker'` or `type: 'task:rejected'` publishes anywhere in production code on main. Bridge was dormant — Sam's gap claim is real.

### Findings against the diff

**Two new private publishers — ✅ exact mirror of `publishWorkItemQueued` (PR #363)**

Both `publishTaskDoneByWorker` and `publishTaskRejected` follow the canonical pattern established for `workitem:queued`:
- No-op + debug log when EventBus is not wired (test/CLI path).
- Try/catch around `bus.publish()` with warn-log; storage commit is the source of truth.
- Deterministic event id `${type}:${workItemId}` for per-WI dedup.
- Empty `sessionName` (system-level event, not session-attributed) — JSDoc explicitly explains.
- Carries `workItemId` (mandatory for bridge resolution) + `requestId` + `missionId` (correlation).

Verified the bridge consumer side: `event-to-workitem-bridge.service.ts:766` resolves the source via `event.workItemId ?? event.taskId` → `taskPool.findWorkItem(id)`. The new publishers' `workItemId: workItem.id` is exactly what the bridge needs. Wire is intact end-to-end.

**Call site placement — ✅ correct**

- `submitForVerification` (line 668-682): publish AFTER `storage.flush()` AND inside `if (updated)` race-window guard. Subscribers re-reading via `findWorkItem` see the committed `done_by_worker` status.
- `verifyItem` (line 771-779): publish ONLY on `if (verdict === 'rejected' && updated)`. The verified branch correctly stays bridge-silent — terminal-success goes through `resolveBlockedDependents` which is not a bridge trigger. JSDoc explicitly calls this out.

**State machine assumptions — ✅ correct**

- `previousValue: 'running'` for `task:done_by_worker` — only valid transition into `done_by_worker` is from `running`, so unconditional value is safe.
- `previousValue: 'done_by_worker'` for `task:rejected` — only valid transition into `rejected` is from `done_by_worker`, so unconditional value is safe.

Both transitions are guarded by `transitionStatus`, so reaching the publish line means the state machine accepted the transition.

**Test coverage — ✅ comprehensive 4×2 + verified-branch matrix**

9 tests total, dropped into the existing `submitForVerification` and `verifyItem` describe blocks per CLAUDE.md co-location:

`task:done_by_worker publish (4 tests)`:
- Happy path: 1 publish, all fields correct (workItemId / requestId / missionId / prev=running / new=done_by_worker / deterministic id)
- State-machine throw: skip claimFromPool so submitForVerification on `queued` throws → no publish ✓
- No-bus: graceful no-op ✓
- Bus-throws: state commit wins (status === 'done_by_worker' assertion) ✓

`task:rejected publish (5 tests)`:
- Happy path: 1 publish, all fields correct ✓
- **Verified-branch negative test**: `verifyItem(verified)` → no publish (success path bridge-silent) ✓ ← this is the load-bearing extra spec for this publisher's branch discrimination
- State-machine throw ✓
- No-bus ✓
- Bus-throws ✓

The verified-branch negative test is the right extra-spec for this publisher (which has an additional verdict-discrimination branch the other publisher doesn't have).

**Verified locally on PR head `f97ef646`:**
- `npx jest backend/src/services/task-pool/task-pool.service.test.ts` → **93 passed, 93 total** ✓ (matches PR claim of 84+9)
- `npx jest backend/src/services/event-bus` → **77 passed, 77 total** ✓ (zero bridge regression — handlers still resolve correctly with the new publishers)

**1) Spec deviation — F1.A/F1.B prescribed `verification.service.ts` location.** Sam notes this in Out-of-scope. The actual code lives in `task-pool.service.ts` because that's where the state-machine transitions happen. **Splitting would require pulling state-machine knowledge into a thin wrapper service** (more brittle, more indirection, harder to keep transition-and-emission paired). Sam's pragmatic call is correct — the publish site MUST be co-located with the transition, otherwise the coupling moves from "in one file" to "across files via interface contracts." The original spec was wrong; the implementation is right.

**2) The "no-bus" tests assert `resolves.not.toBeNull()` rather than asserting publish-call-count.** This is slightly less explicit than the workitem:queued counterpart in PR #363 (which asserts `publishCalls.length === 0`). Functionally equivalent — without a fakeBus there's no count to assert against. Acceptable but a `let publishedSomething = false; (no fakeBus assigned)` style would have been one extra line of belt. Not blocking.

**3) Future bridge-integration smoke test.** The bridge tests (77/77) confirm the handlers work in isolation, and these 9 tests confirm the publishers fire. There's no end-to-end "publisher → bridge → new WI in pool" smoke test in this PR. That's the right scope cut for now (each side is independently tested), but a follow-up integration spec in a future PR would close the loop and catch any bridge-handler ↔ publisher contract drift. Worth filing as a follow-up, not blocking.

### Process notes

- Re-fetched PR body fresh (length 5594 stable).
- Verified the gap claim with a fresh grep on main pre-PR — confirmed zero production publishers existed.
- Verified the bridge consumer wire by reading `event-to-workitem-bridge.service.ts:766` — `event.workItemId ?? event.taskId` → `findWorkItem`. Publishers correctly set `workItemId: workItem.id`.
- Worktree pattern in use (`/tmp/crewly-f1-publishers-sam`) — fifth PR validating SOP from #362.

✅ Ready to merge. Real-functional-gap fix shipped. Bridge handlers are no longer dormant — the autonomy chain (worker submit → TL verify → retry/escalate) now runs through the bridge in production, not log-grep. Good TL judgment from Sam to prioritize this.

PR #382, opened 2026-04-30
fix(autonomy_v1): Request stuck on Active — SLA close uses legal WI transitions + cascades parent Request

stevehuang0115:
## Summary

- Fixes Steve-reported bug (Slack `D0AC7NF5N7L/1777515218.226079`): /tasks UI showed Active=56, Done=41 with many already-replied Requests stuck on Active. Two compounding bugs masked by a lenient test fake.
- `RequestSlaSubscriber.markResolved` now picks a legal terminal status from `WORK_ITEM_TRANSITIONS` (queued→cancelled, running→done, done_by_worker→verified) instead of always firing `queued → done` and silently throwing.
- After the WI transition, the SLA subscriber **cascades the parent Request to `done`** via `maybeCloseRequest`, gated on "no other non-terminal WIs for this Request" so an orc that decomposed + replied doesn't prematurely close a Request whose delegate WIs are still running.
- Test fake hardened to enforce real `WORK_ITEM_TRANSITIONS` — the same illegal transition that broke prod now fails the unit test class.
- Frontend `mapRequestStatus` lists every backend status explicitly (replaces stale `'in_progress'` literal); unknown statuses now `console.warn` instead of silently rendering as Active.

## Root cause (1-line)

`transitionStatus(queued → done)` is illegal per `WORK_ITEM_TRANSITIONS`; the throw was swallowed at warn level, so the SLA WI stayed `queued` and the parent Request never closed.

## Why this couldn't happen sooner

The fake `TaskPoolService` in `request-sla.subscriber.test.ts` had a lenient `transitionStatus` that skipped state-machine validation. 72 unit tests passed against an impossible production transition. PR includes a hardened fake + 11 new pure-helper tests with property-style assertions that walk every live WorkItemStatus through the helper output and verify the result is reachable per `WORK_ITEM_TRANSITIONS`.

## Trade-off — direct Request update vs publish-from-transitionStatus

Steve's plan-rewrite suggested making `transitionStatus` publish `task:completed` so the existing `cascadeRequestStatus` chain in `v3-data.service` runs automatically. This PR uses direct `RequestService.update` from the SLA subscriber instead because:

- Smallest blast radius — only `request-sla.subscriber.ts` + tests touch
- Avoids orphan running WIs from a queued→running→done two-step (reconciler could pick up the intermediate state)
- No re-entry concerns in the cascade event chain
- The SLA subscriber already owns the inbound user request lifecycle

The publish-from-transitionStatus refactor remains a good architectural follow-up — filing as a separate ticket so this hot fix can ship today.

## Test plan
- [x] `npx jest backend/src/services/v3/request-sla.subscriber.test.ts` — 87 / 87 pass (was 72)
- [x] `npx tsc --noEmit -p frontend/tsconfig.json` — clean
- [x] `npm run build` — clean (no errors / warnings)
- [x] `npx vitest run frontend/src/components/RequestTracking/RequestList.test.tsx` — 10 / 10 pass (was 8; +2 new mapper tests)
- [x] Verify pre-existing failures (`request-notification.service.test.ts`, `escalation.service.test.ts`, `RequestSummaryBar.test.tsx`, etc.) are unchanged on this branch — confirmed via git stash baseline

## Manual repro path (post-deploy)
1. Send a Slack message to the orc that does NOT decompose into delegate WIs (e.g. a question).
2. Wait for the orc to reply in-thread.
3. Reload the /tasks page — the Request should appear under "Done", not "Active".
4. The respond_to_user WI in the pool should be `cancelled` with `metadata.slaResolvedReason='orc_reply'`.
5. Repeat for chat-v2: same outcome with reason `chatv2_reply`.

### DIFF
--- a/backend/src/services/v3/request-sla.subscriber.test.ts
+++ b/backend/src/services/v3/request-sla.subscriber.test.ts
@@ -28,10 +28,17 @@ import {
   setRequestSlaSubscriber,
   getRequestSlaSubscriber,
   respondToUserWorkItemId,
+  pickResolveTarget,
+  pickFailTarget,
+  closeRequestPath,
 } from './request-sla.subscriber.js';
 import type { EscalationSlackCallback } from './request-sla.subscriber.js';
-import type { Request } from '../../types/v2/request.types.js';
-import type { WorkItem, WorkItemStatus } from '../../types/v2/work-item.types.js';
+import type { Request, RequestStatus } from '../../types/v2/request.types.js';
+import {
+  type WorkItem,
+  type WorkItemStatus,
+  WORK_ITEM_TRANSITIONS,
+} from '../../types/v2/work-item.types.js';
 import type { TaskPoolService } from '../task-pool/task-pool.service.js';
 import type { RequestService } from './request.service.js';
 
@@ -79,6 +86,14 @@ function buildRequest(overrides: Partial<Request> = {}): Request {
  * Stub TaskPoolService — captures addToPool + transitionStatus calls and
  * lets each test set the WI's status for findWorkItem so terminal-status
  * branches can be exercised.
+ *
+ * IMPORTANT (2026-04-29 hardening, Steve bug): the fake's `transitionStatus`
+ * now enforces the real `WORK_ITEM_TRANSITIONS` matrix. The previous lenient
+ * stub silently accepted illegal transitions (e.g. `queued → done`),
+ * masking a production-only failure where the SLA close path threw and got
+ * swallowed by `markResolved`'s catch. Using the canonical matrix here
+ * means the same code path tests run against the same gate prod runs
+ * against — that class of bug fails CI next time, not in users' faces.
  */
 function buildFakeTaskPool(): {
   taskPool: TaskPoolService;
@@ -97,6 +112,7 @@ function buildFakeTaskPool(): {
       addCalls.push(wi);
     }),
     findWorkItem: jest.fn(async (id: string) => stored.get(id) ?? null),
+    getAllItems: jest.fn(async () => Array.from(stored.values())),
     transitionStatus: jest.fn(
       async (
         id: string,
@@ -106,6 +122,13 @@ function buildFakeTaskPool(): {
       ) => {
         const wi = stored.get(id);
         if (!wi) return null;
+        // Enforce the real state machine — mirrors TaskPoolService.transitionStatus.
+        const allowed = WORK_ITEM_TRANSITIONS[wi.status];
+        if (!allowed.has(status)) {
+          throw new Error(
+            `Invalid status transition for WorkItem ${id}: ${wi.status} → ${status}`,
+          );
+        }
         wi.status = status;
         if (mutator) mutator(wi);
         transitionCalls.push({ id, status, actor });
@@ -126,17 +149,35 @@ function buildFakeTaskPool(): {
 }
 
 /**
- * Stub RequestService — only `getById` is touched.
+ * Stub RequestService — `getById` for read, `update` for the cascade close
+ * path added in 2026-04-29's bug fix. The stub mutates the in-memory
+ * registry and records every status change so tests can assert the
+ * Request lifecycle without spinning up filesystem persistence.
  */
 function buildFakeRequestService(initial: Request[] = []): {
   service: RequestService;
   registry: Map<string, Request>;
+  updateCalls: Array<{ id: string; status?: RequestStatus; result?: string }>;
 } {
   const registry = new Map<string, Request>(initial.map((r) => [r.id, r]));
+  const updateCalls: Array<{ id: string; status?: RequestStatus; result?: string }> = [];
   const service = {
     getById: jest.fn(async (id: string) => registry.get(id) ?? null),
+    update: jest.fn(async (id: string, updates: Partial<Request>) => {
+      const r = registry.get(id);
+      if (!r) throw new Error(`Request not found: ${id}`);
+      if (updates.status !== undefined) r.status = updates.status;
+      if (updates.result !== undefined) r.result = updates.result;
+      r.updatedAt = new Date().toISOString();
+      updateCalls.push({
+        id,
+        status: updates.status,
+        result: updates.result,
+      });
+      return r;
+    }),
   } as unknown as RequestService;
-  return { service, registry };
+  return { service, registry, updateCalls };
 }
 
 /**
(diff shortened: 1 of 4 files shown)


stevehuang0115 (commented):
# Arch review — verdict: APPROVE ✅

(Posted as a comment review because GitHub blocks self-approve; treat this as my arch sign-off.)

Solid scoped fix. State-machine paths are correct, the fake hardening is the right shape, and the trade-off call is sound. Items below are nice-to-haves / follow-up scope, not blockers.

---

## Verdict on Steve's review questions

### 1. State-machine transition paths — correct

Walked the helpers against `WORK_ITEM_TRANSITIONS` and `REQUEST_TRANSITIONS` directly:

- `pickResolveTarget`: `running→done`, `done_by_worker→verified`, everything else `→cancelled` — every output is a legal edge from the input. Property test on lines 252-271 pins this against future drift.
- `pickFailTarget`: `running→failed`, `done_by_worker→rejected`, everything else `→cancelled` — same property guarantee.
- `closeRequestPath`: `open/running/waiting_confirmation→done` direct, `ready/blocked→running→done` two-step — matches `REQUEST_TRANSITIONS` exactly. Two-step is safe because `RequestService.update` re-validates each hop (request.service.ts:270).

### 2. Other silent-catch sites — none in production code

Audited every `transitionStatus` call site in `backend/src/` (excluding tests):

| Caller | From-status guard | Verdict |
|---|---|---|
| `task-pool.service.ts:449` claimFromPool | candidate filtered by `status === 'queued'` | safe |
| `task-pool.service.ts:512` claimSpecificItem | `workItem.status !== 'queued'` early return | safe |
| `task-pool.service.ts:570` releaseBack | `running`/`blocked` guard at line 547 | safe |
| `task-pool.service.ts:660` submitForVerification | `running` guard | safe |
| `task-pool.service.ts:713` completeSimpleItem | `running` guard | safe |
| `task-pool.service.ts:761` verifyItem | `done_by_worker` guard | safe |
| `task-pool.service.ts:854` resolveBlockedDependents | filter by `status === 'blocked'` | safe |
| `task-pool.service.ts:894` failItem | `running` guard at line 875 | safe |
| `request-sla.subscriber.ts:614` markResolved | **was the bug** | now uses `pickResolveTarget` ✓ |
| `request-sla.subscriber.ts:1060` failOrphanRespondWi | **was also buggy** | now uses `pickFailTarget` ✓ |

Production code is clean of this class of bug after this PR.

### 3. Test fake hardening — sufficient

The fake at lines 116-137 enforces `WORK_ITEM_TRANSITIONS` exactly the way `TaskPoolService.transitionStatus` does. Error message even mirrors the prod string. Property tests at lines 252-271 / 287-303 walk every live status through both helpers and assert the result is a legal edge — same drift that broke prod now fails CI loudly.

One follow-on (post-merge, see N1 below): the `RequestService` fake's `update` (lines 166-178) does NOT enforce `REQUEST_TRANSITIONS`. If a future `closeRequestPath` change ever produced an illegal Request transition, it would silently pass tests while prod would throw.

### 4. cascadeRequestStatus alternate paths — covered

Audited the cascade trigger graph:

- `v3:task_completed` / `v3:task_failed` / `v3:task_blocked` → v3-data.service handlers → `cascadeRequestStatus`. These all fire from the file-based task system (controllers), not from `transitionStatus`. **Sam's path doesn't double-cascade — direct update only.**
- Orc-decomposed-into-child-WIs case is handled correctly: when child WIs complete via `task:completed`, the file-layer cascade fires and `cascadeRequestStatus` closes the Request. The SLA subscriber's `countOtherActiveWorkItems` gate at line 671 explicitly defers to that path.
- One asymmetry: `failOrphanRespondWi` does NOT call `maybeCloseRequest`. That's intentional (10-min escalation = user gave up, Request stays in flight for orc to handle) but worth a code comment — see N3.

---

## On the trade-off (publish-from-transitionStatus, deferred)

**Sam's call is right, and there's a stronger reason than the PR body spells out.**

`cascadeRequestStatus` at v3-data.service.ts:725 has this rule:

```ts
} else if (statuses.every((s) => s === 'failed' || s === 'cancelled')) {
  newStatus = 'cancelled';
}
```

If publish-from-transitionStatus were wired in as Steve's plan-rewrite suggested, the SLA cancel path (queued→cancelled on a respond_to_user WI that is the ONLY WI for the Request) would cascade to **`Request = cancelled`, not `done`** — semantically wrong. The cancel here means "no longer needed because the orc handled this", not "abandoned".

So the deferred refactor as filed is **not trivially valuable**. If we do file it, the ticket needs to state explicitly that `cascadeRequestStatus`'s all-failed-or-cancelled rule must ALSO change for inbound Requests — otherwise wiring publish into transitionStatus would actively regress this hot fix.

→ See N2 below — recommend re-scoping the follow-up.

My read: status quo is fine. The SLA subscriber owns inbound Request lifecycle by domain; co-locating the close call there is a feature, not a layering violation.

---

## Suggestions (post-merge, all non-blocking)

### N1 — Mirror REQUEST_TRANSITIONS enforcement in the test fake
The WorkItem fake now enforces the matrix; the Request fake doesn't. Add `isValidRequestTransition` check at `request-sla.subscriber.test.ts:166-178` so a future drift in `closeRequestPath` fails CI before it reaches prod.

### N2 — Re-scope the publish-from-transitionStatus follow-up
The ticket as filed misses that `cascadeRequestStatus`'s all-cancelled rule would interact badly. Either re-title to capture both halves ("decide where Request close authority lives for inbound Requests"), or close the follow-up as "decided not to pursue" and document the SLA-subscriber-owns-close decision in v3 architecture notes.

### N3 — One-line comment on the markResolved/failOrphanRespondWi asymmetry
At `request-sla.subscriber.ts:1045`, add a comment confirming the no-cascade is intentional ("10-min escalation = user gave up; Request stays in-flight"). The kind of asymmetry future refactors like to "fix" by accident.

### N4 — Minor PR-body doc-drift
PR body claims "72 → 87 unit tests (15 new)". Actual count is 53 → 72 (+19 new). Tests pass and substance is correct — this is just a counting drift. Flagging for next time, no action needed if you've already merged.

### N5 — Frontend mapper exhaustiveness claim
The doc comment at `RequestList.tsx:56` says "the exhaustiveness assertion at the bottom is the guard". There's no actual TS exhaustiveness check (`assertNever`); the runtime `console.warn` is the actual guard. Either fix the doc, or narrow `backendStatus` to `RequestStatus` and add `const _exhaustive: never = backendStatus` to make the claim true.

---

## Test verification on my end

- `npx jest backend/src/services/v3/request-sla.subscriber.test.ts` → **72 / 72 pass** (matches actual file content)
- `cd frontend && npx vitest run src/components/RequestTracking/RequestList.test.tsx` → **10 / 10 pass**
- `npx tsc --noEmit -p backend/tsconfig.json` → clean
- `npx tsc --noEmit -p frontend/tsconfig.json` → clean

Ship it. 🚀

— Architect (crewly-product-arch-fe8119fd)

PR #379, opened 2026-04-30
fix(browser-ext): TTL sweep + dedup + threshold-based status UI

stevehuang0115:
## Summary

Cloud Portal Browser Extensions panel was showing **every** ext instance as green "Online" forever, even when `lastSeenAt` was 30+ minutes stale. Reinstalling the extension also produced phantom rows alongside the live one — the relay's `disconnected` event is occasionally lost and the backend Map had no TTL.

This PR fixes both ends:

- **FE (`029865f7`):** new `getInstanceStatus(lastSeenAt)` helper drives a 3-state badge (Online ≤30s · Stale ≤60s · Offline >60s or missing). CloudPortal renders amber / gray instead of always-green.
- **BE (`b7ff624d`):** `BrowserProxyService` gains a wall-clock TTL sweep (every 60s, purges entries >5 min stale) and an optional `deviceFingerprint` field that collapses re-install ghost rows on register. Backward compatible — when fingerprint is absent (older ext build), dedup is a no-op and the TTL sweep alone still cleans things up over the 5-min boundary.

Spec: `.crewly/specs/browser-ext-stale-status-fix-2026-04-30.md` (local; gitignored).

### Frontend
- [x] `lastSeenAt` ≤ 30s ago → **Online** (emerald)
- [x] 30s < lastSeenAt ≤ 60s → **Stale** (amber)
- [x] lastSeenAt > 60s OR missing → **Offline** (gray)
- [x] Both color and text label change

### Backend
- [x] (a) Every 60s, sweep `instances` Map; delete entries where `(now - lastSeenAt) > 5 min`. Logs each purge with `instanceId` + `instanceName`
- [x] (b) On register (`handleBrowserList` rebuild OR `browser_event:connected`): if incoming has `deviceFingerprint` and an existing entry shares it, delete the old. Backward compatible: empty/absent fingerprint → no-op
- [x] (c) Unit tests cover (a) + (b) — actually 9 new BE tests + 13 new FE tests

### Cross-cutting
- [x] No magic numbers — `BROWSER_EXT_STATUS` (FE) and `BROWSER_PROXY_CONSTANTS` (BE) namespaces with `as const`
- [x] 1:1 source-test ratio for every new file
- [x] JSDoc on every new function/method/type
- [x] No touched files outside spec scope (Steve's WIP for per-tab binding / Knowledge V2 / BrowserExtensionPanel was surgically stashed before work began — see `stash@{0}: WIP-foreign 2026-04-30 03:30Z park-by-Sam`)

[...]

PR #383, opened 2026-04-30
fix(browser-proxy): periodic list_browsers refresh prevents live ext eviction

stevehuang0115:
## Summary

Fixes Steve's bug from 2026-04-30: Chrome ext shows **Connected** via Cloud Relay (91ms RTT, heartbeats flowing) but OSS backend `/api/browser/instances` returns `[]`.

This is a regression introduced by #379. The TTL sweep added there evicts live extensions because the `lastSeenAt` field never advances for an idle-but-connected ext (no tool calls = no relay events = frozen timestamp). 5 min after registration, the sweep purges the live entry.

**Smoking gun from `~/.crewly/logs/crewly-2026-04-30.log`:**
```
02:09:38  INFO  [BrowserProxy] Browser instance connected | count: 1
02:15:27  INFO  [BrowserProxy] Purged stale browser instance | lastSeenAt: 02:09:38
```
Exactly the 5-min mark, and the ext is still heartbeating to the relay the whole time.

## Fix

Add a periodic `list_browsers` refresh (60s cadence) so the proxy actively re-asks the relay for the current set. The relay knows the ext is alive (it's receiving heartbeats from it) and returns a fresh `lastSeenAt`, which `handleBrowserList` propagates onto the in-memory entry. Refresh interval (60s) is strictly less than the purge threshold (5 min), giving a 5× safety margin against transient relay slowdowns.

- New constant: `BROWSER_PROXY_CONSTANTS.LIST_BROWSERS_REFRESH_INTERVAL_MS = 60_000`
- New private fields/methods: `refreshTimer`, `startRefreshTimer()`, `stopRefreshTimer()`
- Symmetric lifecycle with `sweepTimer`: starts from `case 'registered'`, stops in both `handleDisconnect` and `cleanup` (no leaked intervals on reconnect)

## Tests (3 new, 50 total passing on this file)

- `sends list_browsers periodically after registration` — counts JSON-decoded `list_browsers` payloads on the mock WS
- `keeps a live extension visible past STALE_PURGE_THRESHOLD_MS when relay refreshes lastSeenAt` — drives 6 min of fake-time in 60s steps, simulating relay's snapshot response each tick
- `disconnect clears the refresh timer (no leaked interval after teardown)` — asserts no additional `list_browsers` sends after `disconnect()` across multiple intervals

## Test plan

- [x] `npx jest backend/src/services/browser/browser-proxy.service.test.ts` — 50/50 pass
- [x] Confirmed pre-existing cloud-test failures (`cloud-sync.service`, `cloud-initializer`, `cloud-connect-e2e`) reproduce identically on `main` without these changes — unrelated regression to file separately if not already tracked
- [ ] After merge: restart local OSS backend on Steve's machine; verify `/api/browser/instances` shows the live ext for >5 min without bouncing

## Workaround until merge

Restart the OSS backend — the proxy re-registers, `list_browsers` populates the map, and the ext stays visible for the next 5 min before re-eviction.

### DIFF
--- a/backend/src/constants.ts
+++ b/backend/src/constants.ts
@@ -122,6 +122,15 @@ export const BROWSER_PROXY_CONSTANTS = {
 	 *  relay/network blips without prematurely evicting a live ext connection
 	 *  (heartbeat fires every 25s, so 5 min = ~12 missed heartbeats). */
 	STALE_PURGE_THRESHOLD_MS: 5 * 60_000,
+	/** Cadence (ms) at which the proxy actively re-requests the relay's full
+	 *  `browser_list` so each live instance's `lastSeenAt` advances before the
+	 *  TTL sweep can mistakenly purge it. Without this active refresh, an
+	 *  idle-but-connected extension (no tool calls in flight) sees its
+	 *  `lastSeenAt` frozen at the moment of the initial `connected` event,
+	 *  and the 5-min sweep evicts the live entry. The refresh interval must
+	 *  be strictly less than `STALE_PURGE_THRESHOLD_MS` — 60s gives a 5×
+	 *  safety margin against transient relay slowdowns. */
+	LIST_BROWSERS_REFRESH_INTERVAL_MS: 60_000,
 } as const;
 
 // Agent runtime types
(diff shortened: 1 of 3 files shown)


stevehuang0115:
Closing as superseded by the heartbeat-tick approach now in main (see #392's logic in browser-proxy.service.ts). This timer/setInterval version solves the same idle-eviction problem but is no longer needed.

PR #386, opened 2026-04-30
fix(terminal-gateway): bounded retries close eager-monitoring race (Bug 2 / orc terminal queue delivery)

stevehuang0115:
## Summary

- **RCA reframe:** Steve's "messages queued and never delivered to orc terminal" is an **OUTPUT visibility** bug, not an input-delivery one. Dispatcher (`QueueProcessorService → AgentRegistrationService.sendMessageToAgent → session.write`) is independent of subscribers and was working. The bug lives in the OUTPUT path: when no client is subscribed, `terminal.gateway.startPtyStreaming` pipes bytes through `io.to('terminal_<session>').emit(...)` which is a silent no-op — bytes are lost in the gateway and survive only in the headless-terminal buffer. Opening the panel triggers `sendCurrentTerminalState` which drains the buffer in one burst → "all messages flushed at once."
- **Spec gap:** `.crewly/specs/rca-readiness-state-machines-2026-04-29.md` recommended Option 1 (eager `startOrchestratorChatMonitoring` at boot/setup/restart). That wiring is **already on `main`** at `index.ts:2052`, `orchestrator.controller.ts:426`, `orchestrator-restart.service.ts:257`. The spec's premise — "function exists but is never invoked" — is outdated.
- **Residual hole this PR closes:** the eager call fires the moment `createAgentSession` resolves, but `SessionBackend` may not have registered the new PTY by then. `startPtyStreaming` returns `false` → `persistentMonitoringSessions.delete()` runs synchronously → **no retry** → no `onData` listener for the entire orc lifetime, until a client subscribes.

## Patch

`backend/src/websocket/terminal.gateway.ts` (~80 LOC source, ~120 LOC tests):

1. `startOrchestratorChatMonitoring` schedules bounded retries on a `100/500/1000/2000/5000ms` backoff (≈8.6s budget, covers typical sub-1s backend-register lag with headroom for ~3s heavy-load case). Backoffs live in `TERMINAL_GATEWAY_CONSTANTS.MONITORING_RETRY_BACKOFFS_MS`.
2. `persistentMonitoringSessions` stays flagged across the retry window so a parallel `subscribeToSession` path doesn't bypass eventual attach.
3. `stopOrchestratorChatMonitoring` and `destroy()` clear pending retry timers so a queued retry never fires after the caller gave up.
4. Timers are `.unref()`'d so they never keep the node process alive past graceful shutdown.
5. Calling `start` twice replaces the stale schedule (no double-stacking).
6. On full retry exhaustion: `ERROR` log with `backend.listSessions()` for diagnostics; persistent flag dropped so subscriber-driven attach can still proceed.

## Out of scope (separate follow-ups)

These shipped in Steve's original cluster but are **different code paths** — flagged for their own tickets, NOT included here per the "only modify terminal queue delivery" constraint:

- (B) `delegate-task → "Failed to create WorkItem in TaskPool"` — `TaskPoolService.addToPool` path.
- (C) heartbeat `queuePending` / `queueProcessing = "?"` — UI exposure, fields not surfaced by current heartbeat endpoint.

## Test plan

- [x] Targeted: `npx jest --testPathPattern="backend/src/websocket/terminal.gateway.test.ts"` → **46 / 46 pass** (39 prior + 7 new).
- [x] Wider regression: `npx jest --testPathPattern="backend/src/(websocket|services/orchestrator|controllers/orchestrator)"` → **340 / 340 pass** (2 skipped, unrelated to this change).
- [x] `npm run build` → green.
- [x] `npx tsc --noEmit` (backend) → clean.

## New test cases

In `backend/src/websocket/terminal.gateway.test.ts`:
- Sync attach when PTY pre-registered (happy path unchanged).
- First-attempt false + retry queued when PTY missing.
- Attach on later retry once session becomes available.
- Full schedule exhaustion when session never registers.
- Stop mid-backoff cancels in-flight retries (no leaks).
- Destroy mid-backoff clears retry timers (no leaks).
- Double-start replaces stale schedule (no double-stacking).

## Acceptance criteria for Steve

- After this lands: `send-message → orc` continues writing into the orc PTY as before (no input-path change), AND orc's stdout streams to subscribers from the moment the eager monitor attaches — even if the user never opens the side panel for the rest of the orc's lifetime.
- "Burst on panel open" symptom should disappear because there's no longer a window where orc output goes to /dev/null.

### DIFF
--- a/backend/src/constants.ts
+++ b/backend/src/constants.ts
@@ -194,6 +194,22 @@ export const TERMINAL_GATEWAY_CONSTANTS = {
 	MAX_BUFFER_SIZE: 100 * 1024,
 	/** Trim threshold for partial line buffer (characters) */
 	BUFFER_TRIM_THRESHOLD: 1000,
+	/**
+	 * Backoff schedule (ms) for retrying eager orchestrator chat monitoring
+	 * when the PTY session is not yet registered in the SessionBackend.
+	 *
+	 * Background: startOrchestratorChatMonitoring is invoked from boot,
+	 * setup, and restart paths immediately after createAgentSession resolves.
+	 * The SessionBackend may not have fully registered the new session by
+	 * then — leaving the orchestrator's PTY without an onData listener for
+	 * the rest of its lifetime, which silently drops responses to user
+	 * messages until the side panel is opened (RCA 2026-04-29 / 2026-04-30).
+	 *
+	 * Total budget ≈ 8.6s with 5 attempts. Tuned to outlast typical
+	 * createAgentSession→backend-register lag (sub-1s on local, up to ~3s
+	 * under heavy load) without unbounded retries.
+	 */
+	MONITORING_RETRY_BACKOFFS_MS: [100, 500, 1000, 2000, 5000] as readonly number[],
 } as const;
 
 // Chat routing constants (message markers and patterns for orchestrator communication)
--- a/backend/src/websocket/terminal.gateway.test.ts
+++ b/backend/src/websocket/terminal.gateway.test.ts
@@ -554,4 +554,121 @@ describe('TerminalGateway', () => {
 			expect(mockLogBuffer.removeListener).toHaveBeenCalledWith('data', expect.any(Function));
 		});
 	});
+
+	describe('startOrchestratorChatMonitoring (eager-attach race fix, RCA 2026-04-30)', () => {
+		const orcSession = 'crewly-orc';
+
+		beforeEach(() => {
+			jest.useFakeTimers();
+		});
+
+		afterEach(() => {
+			jest.useRealTimers();
+		});
+
+		it('attaches synchronously when the PTY is already registered', () => {
+			mockSessionBackend.getSession.mockReturnValue(mockSession);
+
+			const result = gateway.startOrchestratorChatMonitoring(orcSession);
+
+			expect(result).toBe(true);
+			expect(mockSession.onData).toHaveBeenCalledTimes(1);
+			expect(gateway.getConnectionStats().activeStreams).toBe(1);
+		});
+
+		it('returns false and schedules a retry when the PTY is not yet registered', () => {
+			mockSessionBackend.getSession.mockReturnValueOnce(null);
+
+			const result = gateway.startOrchestratorChatMonitoring(orcSession);
+
+			expect(result).toBe(false);
+			// Listener not attached on first attempt — backend hadn't registered the session.
+			expect(mockSession.onData).not.toHaveBeenCalled();
+			// Persistent flag must remain set across the retry window so a parallel
+			// subscriber doesn't bypass eventual attachment.
+			expect(gateway.getConnectionStats().activeStreams).toBe(0);
+		});
+
+		it('attaches on a later retry once the session becomes available', () => {
+			// First attempt: session not yet registered. Subsequent attempts: registered.
+			mockSessionBackend.getSession
+				.mockReturnValueOnce(null)
+				.mockReturnValue(mockSession);
+
+			gateway.startOrchestratorChatMonitoring(orcSession);
+			expect(mockSession.onData).not.toHaveBeenCalled();
+
+			// First retry fires after 100ms — should attach.
+			jest.advanceTimersByTime(100);
+
+			expect(mockSession.onData).toHaveBeenCalledTimes(1);
+			expect(gateway.getConnectionStats().activeStreams).toBe(1);
+		});
+
+		it('keeps retrying through the full backoff schedule before giving up', () => {
+			// Backend never registers the session — exhaust all retries.
+			mockSessionBackend.getSession.mockReturnValue(null);
+
+			gateway.startOrchestratorChatMonitoring(orcSession);
+
+			// Total budget = 100 + 500 + 1000 + 2000 + 5000 = 8600ms. Advance past it.
+			jest.advanceTimersByTime(9000);
+
+			// All getSession lookups happened: 1 sync + 5 retries = 6.
+			expect(mockSessionBackend.getSession).toHaveBeenCalledTimes(6);
+			expect(mockSession.onData).not.toHaveBeenCalled();
+			// After exhaustion the persistent flag is dropped so subscriber-driven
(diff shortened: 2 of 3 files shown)


stevehuang0115:
## Architect review — APPROVE (posted as comment due to gh-identity collision; treat as formal sign-off)

**Verdict: 0 must-fix, 1 nice-to-have. Ship.**

### Bounded-retry termination (✅ all paths terminate)

Walked the state machine. The retry loop has **four exits**, all bounded:

1. **Sync attach succeeds** (line 402) → `started=true` returned, no timer scheduled.
2. **Retry attaches** (line 458 `started=true`) → return inside `setTimeout` callback, no `scheduleMonitoringRetry(attemptIndex+1)` call. Map entry already deleted at line 449.
3. **Schedule exhausted** (line 425 `attemptIndex >= schedule.length`) → ERROR log + persistent flag dropped + map entry deleted, no further `setTimeout`.
4. **Stop guard trips** (line 451 `!persistentMonitoringSessions.has`) → debug log, return without rescheduling. Stop happens via `stopOrchestratorChatMonitoring` (which clears the timer first via `clearPendingMonitoringRetry`) or via `destroy()` (which clears all timers in the for-loop at line 1019).

No path recurses without consuming an `attemptIndex`. Total budget = 100+500+1000+2000+5000 = 8600ms hard ceiling. Cannot loop indefinitely. ✅

### New race window introduced by retry? (✅ none)

Three potential race surfaces analyzed:

- **Subscribe-during-retry:** subscriber path calls `subscribeToSession` → `startPtyStreaming` (line 210). `startPtyStreaming` line 279-282 has `if (this.sessionSubscriptions.has(sessionName)) return true` early-return. So subscriber-attaches-first means the retry's later `startPtyStreaming` call sees the existing subscription and returns true (already streaming) — no double-attach, no double `onData` listener. Verified by code-read.

- **Stop-during-retry:** `stopOrchestratorChatMonitoring` calls `clearPendingMonitoringRetry` (line 472, before `stopPtyStreaming`). Even if a timer happens to be firing concurrently, the in-flight callback's stop-guard (line 451 `!persistentMonitoringSessions.has`) catches the case where stop dropped the flag between `setTimeout` resolution and listener attach. Belt-and-suspenders. ✅

- **Destroy-during-retry:** `destroy()` clears all pending timers (line 1018-1023) before tearing down `sessionSubscriptions` and `connectedClients`. Same stop-guard protection on any concurrent in-flight callback. ✅

- **Double-start race:** First call schedules retry. Second call hits `clearPendingMonitoringRetry` at line 388 before scheduling a fresh schedule. Old timer is cancelled, no double-stacking. ✅ Test "replaces a stale retry schedule" locks this.

`.unref()` on every timer (line 466) means lingering retries cannot block graceful shutdown.

### Test coverage for eager-monitor path (✅ solid, 1 nice-to-have)

7 new tests cover the full state machine: sync-attach / first-false-retry-queued / attach-on-retry / full-exhaustion / stop-mid-backoff / destroy-mid-backoff / double-start-replaces-stale. Targeted 46/46 + wider 340/340 green.

**N1 (nice-to-have, NOT blocking):** No explicit test for *parallel-subscribe-during-retry*. The contract is held by `startPtyStreaming`'s idempotency guard (line 279) — verified by code-read, but not locked at the test surface. If subscribe path or retry path ever changes their early-return semantics, this race could regress silently. Suggested ~10-LOC follow-up:

```ts
it('subscribe-during-retry does not double-attach', () => {
  mockSessionBackend.getSession.mockReturnValueOnce(null).mockReturnValue(mockSession);
  gateway.startOrchestratorChatMonitoring(orcSession);   // schedules retry
  gateway.subscribeToSession(orcSession, mockSocket);     // attaches first
  jest.advanceTimersByTime(100);                          // retry fires
  expect(mockSession.onData).toHaveBeenCalledTimes(1);    // not 2
});
```

Lock-in only — no behavior change. Open as a follow-up ticket; do NOT block this merge on it.

### Side-effect check (✅ scoped)

- 3 callers verified fire-and-forget: `index.ts:2052` (boot), `orchestrator.controller.ts:426` (setup), `orchestrator-restart.service.ts:257` (restart). Return-value semantic change (`false` = "retry scheduled" instead of "failed permanently") is safe — no caller branches on the boolean. JSDoc at line 374 updates the contract correctly.
- No chat path / autonomy_v1 path touched. `dispatcher.service.ts` (chat ingress) and `scheduler.service.ts` / `work-item.service.ts` (autonomy_v1) interact with the PTY via `session.write` (input path), independent of the gateway's `onData` listener (output path). Fix is scoped purely to the output path.

### Architecture fit

The "gateway-buffer-when-no-subscriber" alternative noted in the PR body (= RCA Option 3) has **strictly higher blast radius** — every PTY's behavior changes, eviction policy needed, memory pressure if no one ever subscribes. Retry-at-boundary keeps the change scoped to orchestrator monitoring only — agent output capture, learning auto-record subscriber, WorkItem heartbeat reviewer all keep their current PTY contract. Compose-at-boundary discipline upheld: `pendingMonitoringRetries` map is private to the gateway, retries operate on the same `startPtyStreaming` primitive that the subscriber path uses, no new shared mutable state.

### Constants

`MONITORING_RETRY_BACKOFFS_MS = [100, 500, 1000, 2000, 5000]` — total 8.6s, 5 attempts. 100ms first retry covers sub-1s race, cumulative 3.6s by attempt-4 covers documented heavy-load ~3s case, 5s tail is safety margin. Naming consistent with sibling constants under `TERMINAL_GATEWAY_CONSTANTS`. JSDoc cites the RCA + measured race window — future archeologists can trace the design.

### Process verifications

- ✅ Re-fetched PR body before posting (per PR #348 doc-drift discipline) — no drift between fetch-time and review-post.
- ✅ Grepped 3 callers + verified fire-and-forget (per PR #358 grep-for-callers-before-approving discipline).
- ✅ Verified idempotency claim against actual line numbers in the diff (per WIRE-1 verify-don't-trust-claims discipline).
- ✅ Severity discipline: 0 must-fix, 1 nice-to-have, fix shape + LOC cost paired with the nice-to-have.

**Skim+ACK close-out: no re-review needed. Merge when CI is green.**

PR #411, opened 2026-05-03
docs(sops): add Dev Process Tiers SOP (P0 — Decision 3)

stevehuang0115:
**Tier:** Fast
**Why this tier:** New SOP doc, internal-only, loaded on demand via `get-sops`. Agent prompt integration is a separate PR.

## Summary
- Adds `config/sops/common/dev-process-tiers.md` — defines three delivery tiers (Fast / Standard / Release Path) with a decision matrix, default selection rules, and per-tier required deliverables.
- Each PR is asked to declare its tier in the description; reviewers can re-tier.
- No agent prompts touched. No code touched. SOP loads via existing `get-sops` skill.

## Why this PR is small on purpose
The 2026-05-03 P0 plan calls for tier-based dev process. We land the SOP first in isolation so it can be revised without re-rolling any agent prompt. Integration with `delegate-task` / role prompts is a follow-up PR.

## Eval Criteria
- [x] SOP frontmatter parses (id, version, role, category, priority, title, description, triggers, tags)
- [x] Decision matrix covers customer-facing / persistence / money-security / public-release axes
- [x] Default selection rules stated (greenfield → Fast, customer-facing → Standard, money/security/release → Release Path)
- [x] Each tier lists required deliverables and what is skipped vs the next tier up
- [x] Anti-patterns called out (size ≠ tier, splitting Release Path across Fast PRs, etc.)
- [x] No agent prompt mass changes

## Test Plan
- [x] Frontmatter validated (id/title/triggers/tags present, parse-clean)
- [x] `npm run build` clean (no source files touched)
- [ ] `get-sops` retrieves with trigger keywords like "which tier", "dev process", "release path" — to be validated post-merge by Sam

## Refs
- `.crewly/specs/2026-05-03-agent-improvement-p0-execution.md` — Decision 3, deferred-to-P1 in the spec, reclassified P0 by Steve in 2026-05-03 dispatch.
- Companion follow-up will reference this SOP from delegate-task skill and TL role prompts.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #412, opened 2026-05-03
feat(observability): minimal internal metrics — agent_behavior_log (P0 / Decision 4)

stevehuang0115:
## Summary
Lands the **schema + write surface** for Crewly's first-party behavior log — the foundation for the four agent-prompt-improvement P0 metrics agreed in `specs/2026-05-03-agent-improvement-p0-execution.md` (Decision 4):

- `agent.action` — delegate / implement / ask_human / send_slack / review / verify / other
- `agent.escalation` — agent → upstream escalations w/ reason
- `slack.delivery.failed` — Slack messages we intended to deliver but didn't
- `prompt.size.bytes` — per-session compiled-prompt byte size samples

All four schemas land in this PR. Single SQLite table (`agent_behavior_log`) with `event_type` discriminator + flexible row shape (`agent / subject / task_id / reason / amount / details_json`) — additive event types in the future will not need migrations.

## Scope — what's in / what's out

**In this PR (small, reviewable):**
- `services/observability/observability-db.ts` — lazy `better-sqlite3` opener, idempotent migration, indexes for digest queries. Mirrors the proven `chat-v2/sqlite/chat-db.ts` pattern.
- `services/observability/agent-behavior-log.types.ts` — discriminated union of the four event types + storage row shape + type guard.
- `services/observability/agent-behavior-log.service.ts` — typed append-only `record(event)` API + `listRecent` / `countByEvent` for diagnostics. Best-effort writes (never throws into the calling path).
- 1:1 test files per source (per `crewly/CLAUDE.md`).

**Deferred to follow-up PRs (each PR small):**
- **Producer wiring:** `slack.delivery.failed` lands with the P0-5 slack-bridge no-reply detector; `prompt.size.bytes` lands in the prompt-generator follow-up; orchestrator `agent.action` / `agent.escalation` call sites land alongside their respective prompt updates.
- **Aggregation / digest queries:** this PR ships only the helpers needed for diagnostics. The weekly-digest query layer is a separate PR.

## Tier
**Tier:** Fast _(per `config/sops/common/dev-process-tiers.md` — internal-only, additive new service, zero customer-facing surface, no auth/billing impact, low blast radius.)_

## Test plan
- [x] Types unit test (4/4) green locally
- [x] Backend `tsc --noEmit` clean (exit 0, zero diagnostics)
- [ ] CI runs the SQLite-backed tests against an arm64-compiled `better-sqlite3` (this Mac's pre-built binding is x86_64 — same env constraint that excludes `vector-store.service.test.ts`; the `chat-v2/sqlite/chat-db.test.ts` peer follows the same pattern)
- [ ] Reviewer sanity-check on the migration DDL and discriminator branch in `toRowParams`

## Notes
- DB path: `~/.crewly/observability.db` — sits next to `chat.db` so operators have one directory to back up / inspect.
- `record()` is intentionally fire-and-forget; failures are logged, not thrown.
- `toRowParams` uses an exhaustiveness `_never: never` guard so new event types force a TypeScript error until they're handled.

### DIFF
--- a/backend/src/services/observability/agent-behavior-log.service.test.ts
+++ b/backend/src/services/observability/agent-behavior-log.service.test.ts
@@ -1,0 +1,236 @@
+/**
+ * Tests for `AgentBehaviorLogService`.
+ *
+ * Uses an in-memory SQLite DB so the suite stays hermetic and fast.
+ *
+ * @module services/observability/agent-behavior-log.service.test
+ */
+
+import {
+  AgentBehaviorLogService,
+} from './agent-behavior-log.service.js';
+import {
+  openObservabilityDatabase,
+  type ObservabilityDatabase,
+} from './observability-db.js';
+import { BEHAVIOR_LOG_EVENT_TYPES } from './agent-behavior-log.types.js';
+
+describe('AgentBehaviorLogService', () => {
+  let db: ObservabilityDatabase;
+  let svc: AgentBehaviorLogService;
+  let nowCounter: number;
+
+  beforeEach(() => {
+    db = openObservabilityDatabase({
+      inMemory: true,
+      skipIntegrityCheck: true,
+    });
+    nowCounter = 1_700_000_000_000;
+    svc = new AgentBehaviorLogService({
+      db,
+      now: () => ++nowCounter,
+    });
+  });
+
+  afterEach(() => {
+    svc.close();
+    db.close();
+  });
+
+  describe('record() — agent.action', () => {
+    it('inserts an action event with reason=actionType and JSON details', () => {
+      const id = svc.record({
+        type: BEHAVIOR_LOG_EVENT_TYPES.AGENT_ACTION,
+        agent: 'crewly-orc',
+        actionType: 'delegate',
+        taskId: 'req-42',
+        details: { recipient: 'sam' },
+      });
+
+      expect(id).toBeGreaterThan(0);
+
+      const rows = svc.listRecent(10);
+      expect(rows).toHaveLength(1);
+      const r = rows[0]!;
+      expect(r.eventType).toBe('agent.action');
+      expect(r.agent).toBe('crewly-orc');
+      expect(r.reason).toBe('delegate');
+      expect(r.taskId).toBe('req-42');
+      expect(r.amount).toBe(0);
+      expect(r.details).toEqual({ recipient: 'sam' });
+    });
+
+    it('handles missing optional fields gracefully', () => {
+      const id = svc.record({
+        type: BEHAVIOR_LOG_EVENT_TYPES.AGENT_ACTION,
+        agent: 'a',
+        actionType: 'implement',
+      });
+      expect(id).toBeGreaterThan(0);
+      const r = svc.listRecent(1)[0]!;
+      expect(r.taskId).toBe('');
+      expect(r.details).toEqual({});
+    });
+  });
+
+  describe('record() — agent.escalation', () => {
+    it('writes escalating agent / target / reason correctly', () => {
+      svc.record({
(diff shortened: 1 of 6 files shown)


stevehuang0115 (commented):
## Verdict: LGTM (non-binding; cannot self-approve)

Clean, focused schema-only PR. Exactly what a "small, reviewable" PR should look like — schema + write surface land together, producer wiring deferred to follow-ups per the spec. Build/types clean, 1:1 source/test ratio respected.

### What's solid

- **Single discriminated table** with `event_type` + flexible row shape (`agent / subject / task_id / reason / amount / details_json`) — additive event types stay migration-free. Right call.
- **Exhaustiveness guard** in `toRowParams` (`const _never: never = event;`) forces TS error on new variants. Good defensive design.
- **Lazy `better-sqlite3` loader** mirrors the proven `chat-v2/sqlite/chat-db.ts` pattern — no native module load at import time.
- **Best-effort writes** — `record()` catches, logs, returns `null`. Observability never takes the request path down. Documented clearly.
- **`Math.max(0, Math.floor(event.bytes))`** clamps fractional/negative bytes defensively.
- **WAL + NORMAL synchronous** PRAGMAs — appropriate for append-only log workload.
- **Three indexes scoped to expected query shapes** (ts; event+ts; agent+event+ts) — matches the digest patterns the spec described.
- **Integrity check on open** with fail-soft semantics — gating startup on a corrupt observability DB would be wrong, and you got that right.
- **27 unit tests** covering happy paths, the four event types, listRecent ordering/filtering, countByEvent grouping + sinceTs cutoff, and the close-then-record resilience case.
- **`now()` injection** for deterministic tests. `db` injection for hermetic in-memory testing.
- **Tier=Fast classification is correct** — internal-only, additive new service, zero customer-facing surface, no auth/billing impact, low blast radius. Matches the new dev-process-tiers SOP.

### Non-blocking observations (consider for follow-up)

1. **No size guardrail on `details_json`** — `stringifyDetails` falls back to `'{}'` only on serialize errors. A misbehaving producer could shove a huge object in `details` and bloat the DB row. Suggest documenting a "callers MUST keep details small (< 1KB)" contract in the JSDoc, or adding a soft byte cap with a truncation marker. Not blocking — producer wiring isn't in this PR — but worth flagging before P0-5 / prompt-generator follow-ups land.

2. **No retention / pruning policy** — table grows unbounded. Probably fine for first few months, but worth a tracking issue. Could be as simple as a daily prune job that drops rows older than N days.

3. **`subject` is a "convenience overload" column** — holds `escalatedTo` / `thread` / `sessionId` depending on event type. The `toRowParams` branches make the mapping explicit and the column comment is honest about it. Just noting this for future maintainers — if the union grows and the overloading gets confusing, the escape hatch is to migrate selected fields out of `details_json`.

4. **`isBehaviorLogEventType` is unused inside the service** — that's intentional (service trusts the type system at insert time, the guard is the read-side contract for arbitrary string queries). The JSDoc actually says this clearly. No action needed.

5. **`rowToRecord` trusts `event_type` from the DB** — if someone bypasses the service and inserts a bad `event_type` string by hand, `listRecent` returns malformed records. Probably YAGNI since the service is the only intended writer, but `isBehaviorLogEventType(row.event_type)` could harden read-side parsing for ~3 LOC.

6. **Reviewer sanity-check on the migration DDL and `toRowParams` discriminator** — done. Both are clean. The `task_id`/`amount`/`details_json` defaults match the per-event-type omissions correctly.

### Test plan

The unchecked CI box for arm64 SQLite is a known env constraint (matches `vector-store.service.test.ts` and `chat-v2/sqlite/chat-db.test.ts`). CI will catch it.

Ready to merge once #411 (Dev Process Tiers SOP) is in — this PR's Tier classification cites that SOP. Order matters.

PR #416, opened 2026-05-03
fix(v3): SLA close-path race — threadIndex.set before awaits + retry on index miss

stevehuang0115:
## Summary
- Fixes the **auto-nudge ×2 production bug** seen on ESTestNode 1.6.4: every inbound Request fires both 5min `request:sla_breached` and 10min `escalation_dm` because the SLA was never closing on orc reply.
- Root cause documented in `.crewly/knowledge/learnings.md:2122` (entry **"AUTO-NUDGE VARIANT 2 ROOT CAUSE (2026-05-03 ESTestNode 1.6.4)"**): SlackBridge fires `markResolvedByThread('')` from `queue-processor.service.ts:690` ~10ms after `request:created`; `handleRequestCreated` was awaiting `requestService.getById` + `taskPool.addToPool(wi)` BEFORE setting `threadIndex` / `chatV2Index`, so the bridge's resolve lookup missed.
- The 5/2 fix that claimed "all 5 SLA close paths were wired" was unit-test-only; the await-ordering bug was hidden by tests that synchronously populated the index in the same tick.

## What changed in `request-sla.subscriber.ts` (+173 / -18)

1. **Critical ordering fix in `handleRequestCreated`**: populate `trackedByRequest` + `threadIndex` + `chatV2Index` **synchronously BEFORE** `await taskPool.addToPool(wi)`. Index is now live by the time the microtask boundary returns control.
2. **`markResolvedByThread` / `markResolvedByChatV2`**: on index miss, schedule a single 250ms `setTimeout` retry (`MARK_RESOLVED_RETRY_MS`). Defensive against any future reordering. Uses `.unref?.()` so a stray retry can't keep the event loop alive.
3. **`markResolved`**: wrap `findWorkItem` with `findWorkItemWithRetry` (5 × 50ms = 250ms, `FIND_WI_MAX_ATTEMPTS` × `FIND_WI_RETRY_INTERVAL_MS`) for the addToPool race; warn-log when the WI never appears.
4. **`addToPool` failure rollback**: try/catch wraps the await; on throw → `clearTimeout` breachTimer + escalationTimer, rollback all in-memory tracking, warn-log, rethrow. Prevents leaked timers if the pool is full / store is down.

## New test coverage (`request-sla.subscriber.test.ts` +233 lines, **82/82 passing**)

New `describe('SLA close-path race fix (2026-05-03 auto-nudge variant 2)')` block:
- ✅ populates threadIndex BEFORE awaiting taskPool.addToPool
- ✅ closes the SLA when `markResolvedByThread` fires while addToPool is still in flight
- ✅ retries `markResolvedByThread` on index miss within 250ms
- ✅ retry no-ops if `request:created` never arrives (defensive)
- ✅ rolls back tracking when addToPool throws (no leaked timers)
- ✅ also retries `markResolvedByChatV2` on chat-v2 channel index miss

## Deploy gap (action required after merge)

⚠️ **ESTestNode 1.6.4 deploy does NOT contain this fix.** Currently auto-nudging every inbound thread ×2. Once this PR merges, ESTestNode needs a follow-up deploy (1.6.5+) to stop the in-prod auto-nudge.

## Scope discipline

This PR contains **only** the 2 SLA files. The dirty `fix/sla-close-path-race` parent branch had unrelated mods to `task-pool.controller`, `browser-bridge`, `knowledge-search/learnings-index`, `SkillsTab`, `delegate-task` script, and a new `BrowserExtensionPanel` — all stashed for separate PRs.

## Test plan
- [x] `npx jest backend/src/services/v3/request-sla.subscriber.test.ts` — 82/82 pass locally
- [x] `tsc --noEmit` clean for both modified files (pre-existing typecheck noise in unrelated test files is not introduced here)
- [ ] Smoke on a fresh ESTestNode-equivalent: post inbound Slack msg → orc reply → confirm no breach/escalation fires at 5/10 min
- [ ] Steve review before merge → cut 1.6.5 → redeploy ESTestNode

### DIFF
--- a/backend/src/services/v3/request-sla.subscriber.test.ts
+++ b/backend/src/services/v3/request-sla.subscriber.test.ts
@@ -31,6 +31,8 @@ import {
   pickResolveTarget,
   pickFailTarget,
   closeRequestPath,
+  MARK_RESOLVED_RETRY_MS,
+  FIND_WI_RETRY_INTERVAL_MS,
 } from './request-sla.subscriber.js';
 import type { EscalationSlackCallback } from './request-sla.subscriber.js';
 import type { Request, RequestStatus } from '../../types/v2/request.types.js';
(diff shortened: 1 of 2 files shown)


stevehuang0115 (commented):
## Architect review — LGTM (non-binding; cannot self-approve)

**Verdict:** Strong, structurally-correct fix with comprehensive test coverage. Would approve if I could.

### Root-cause analysis is sound
Verified the race claim against `backend/src/services/messaging/queue-processor.service.ts:692`: the post-delivery `responseRouter.routeResponse(message, '')` does fire fire-and-forget right after `markCompleted`, well before the SLA subscriber's `await taskPool.addToPool` resolves. So the `request:created` index-population was genuinely losing the race against the bridge's `slackResolve('')` callback. This isn't a hypothetical — the await ordering was the structural defect.

### What I checked
1. **Primary fix (reorder in `handleRequestCreated`)** — populating `trackedByRequest` + `threadIndex` + `chatV2Index` synchronously BEFORE the `addToPool` await is the right surgery. Index becomes visible the moment control yields. ✓
2. **`markResolvedByThread` / `markResolvedByChatV2` retry** — covers the residual ~1ms `await requestService.getById` window. Bounded (single 250ms `setTimeout`), unref'd, and only scheduled when the synchronous lookup misses (no double-fire when the happy-path lookup hits). ✓
3. **`findWorkItemWithRetry` (5 × 50ms = 250ms)** — covers the *complementary* race: bridge's `markResolved` fires AFTER index population but BEFORE `addToPool` persistence completes. With the reorder, this window is now wider, so this retry is necessary, not redundant. ✓
4. **`addToPool` failure rollback** — critical because of the new ordering. Without it, a pool-write failure leaks the breach + escalation timers and fires a 5min/10min `escalation_dm` against a never-persisted WI. The catch correctly clears both timers, deletes from all three maps, warn-logs, and rethrows for the bus error path. ✓
5. **Idempotency under concurrent paths** — traced the scenario where both `markResolvedByThread` (bridge) and the scheduled retry race: `markResolved` deletes from `trackedByRequest` + `threadIndex` synchronously at lines 637-638 before any await, so the retry's `threadIndex.get(threadTs)` returns undefined and bails. No double-resolve. ✓
6. **`clearTimeout` on already-cleared timers** — safe in Node; the rollback path is correct even when `markResolved` already cleared them. ✓
7. **Test coverage** — 6 new tests in `describe('SLA close-path race fix (2026-05-03 auto-nudge variant 2)')`, including the rollback case and the chat-v2 mirror. The "populates threadIndex BEFORE awaiting taskPool.addToPool" test pins the ordering invariant directly — that's the right shape, because it'll fail loudly if a future refactor moves the sets back below the await. 82/82 passing per PR description. ✓

### Code quality
- JSDoc on `MARK_RESOLVED_RETRY_MS`, `FIND_WI_MAX_ATTEMPTS`, `FIND_WI_RETRY_INTERVAL_MS`, `findWorkItemWithRetry`, and the reordered `handleRequestCreated` block explains the *why* and references the incident date — matches our standard ("explain why, not just what").
- Constants properly extracted (no magic numbers in hot paths). Both retry windows align at 250ms, which is intentional and worth the matching comment.
- Inline "CRITICAL ORDERING" comment in `handleRequestCreated` calls out the failure mode explicitly — future-proofing against a well-meaning refactor.

### Non-blocking observations
- `findWorkItemWithRetry` is serial 5 × 50ms — for the cancelled-WI cleanup path this is fine (low concurrency, sub-second worst case), but if you ever need to handle high-fan-in resolve storms, consider a `Promise.race` against a max-wait or a `queueMicrotask`-based attempt-0. Not blocking; current code is correct and safe.
- The `void this.markResolved(...)` inside the `setTimeout` retry is safe because `markResolved`'s outer try/catch swallows everything (including `maybeCloseRequest` failures). Worth a one-line comment near the `void` so a future reader doesn't add a `.catch()` thinking it's unhandled. Optional.
- Scope discipline noted in the description (only the 2 SLA files) is correct — the dirty parent branch's unrelated mods stayed out. Good practice.

### Deploy gap acknowledged
PR description correctly flags ESTestNode 1.6.4 still leaks the auto-nudge ×2 in prod and needs a 1.6.5+ redeploy after merge. That's an ops follow-up, not a code concern.

### Recommendation
Merge after Steve's smoke check on a fresh ESTestNode-equivalent confirms the inbound→reply path no longer fires the 5/10min escalation. The code itself is ready.

— Architect (crewly-architect-arch-57d26b22)

PR #414, opened 2026-05-03
feat(souls): team-leader soul + role-boundaries (P0-1 Phase 1, markdown-only)

stevehuang0115:
## Summary

P0-1 Phase 1 — pure-content (markdown-only) deliverable for the **Sam Role Reframe + team-leader Soul** fix. Splits the spec's combined work into two PRs so Phase 1 can ship with zero runtime risk while Phase 2 (builder + Sam's team config) gets proper test coverage.

- **+103 lines, 2 new files, 0 modified files** — no code, no tests needed (existing souls have no tests)
- **Zero runtime impact** — files are inert until Phase 2 wires them into `SoulModule` / `RoleBoundaryModule`
- **No conflict with Leo's M1 work** in `services/prompt/prompt-generator.service.ts` (different file, different layer)

## What this PR adds

1. **`config/souls/team-leader.md`** (77 lines)
   - TL-primary identity ("you deliver outcomes through delegation, not by personally producing every artifact")
   - Default Operating Mode: decompose → delegate → unblock → verify → report
   - **Self-Implementation Exception Rule** — verbatim from spec P0-2 (AND-of-4 constraint)
   - Mirrors existing `developer.md` / `orchestrator.md` / `researcher.md` structure so `SoulModule.resolveSoul()` archetype lookup (`config/souls/{role}.md`) loads it cleanly once Phase 2 flips Sam's role to `team-leader`

2. **`config/roles/team-leader/role-boundaries.md`** (26 lines)
   - Top-level role-boundary doc the spec calls for in P0-1
   - Distinct from the existing `fragments/role-boundary.md` (consumed by `RoleBoundaryModule.loadRoleFragment`) — the fragment stays in place until Phase 2 net-zero cleanup

## What this PR does NOT do (deferred to Phase 2)

- Modify `SoulModule` / `prompt-builder.service.ts` to load TL soul when `canDelegate=true`
- Flip Sam's team config `"role": "developer"` → `"role": "team-leader"`
- Delete the legacy `tl-addon.md` import path / soft "delegate 70-80%" hints
- Add unit tests for the `canDelegate=true` builder branch

Phase 2 is tracked as task #7 and will open as a separate PR (~150-250 LOC) so runtime changes get isolated review and proper smoke-testing via a dummy team-leader member.

## Test plan

Markdown-only PR — no executable code touched, so no test scripts apply.

- [x] Self-Implementation Exception Rule text matches spec lines 79-90 verbatim (`grep` confirmed)
[...]

PR #417, opened 2026-05-03
feat(prompt): P0-1 Phase 2 — wire team-leader soul in prompt builder (depends on #414)

stevehuang0115:
## Summary
Wires team-leader soul into prompt builder when canDelegate=true, per P0-1 Phase 2 spec.

## Stacking
Depends on #414 (team-leader.md soul markdown). Rebase to main after #414 merge.

## Scope
- prompt-builder.service.ts: load team-leader soul + role-boundaries first when canDelegate=true
- soul.module.ts: TL overlay branch
- role-boundary.module.ts: TL overlay branch
- per-member soul wins over TL overlay (explicit override semantics)
- DUMMY tl-test-member smoke test (no Sam config flip in this PR)

## Tests
- Unit tests on canDelegate=true path
- Dummy member assembly verification

## Out of scope (separate PR)
- Sam team config role flip developer→team-leader (P0-1 Phase 2.5, requires self-restart)

### DIFF
--- a/backend/src/services/ai/prompt-modules/role-boundary.module.test.ts
+++ b/backend/src/services/ai/prompt-modules/role-boundary.module.test.ts
@@ -1,5 +1,11 @@
+import * as fs from 'fs';
+import * as path from 'path';
 import { RoleBoundaryModule } from './role-boundary.module.js';
 import { ModuleConfig } from './prompt-module.interface.js';
+
+jest.mock('fs');
+
+const mockedFs = fs as jest.Mocked<typeof fs>;
 
 /**
  * Tests for the RoleBoundaryModule.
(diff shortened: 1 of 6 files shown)


stevehuang0115 (commented):
## Architect review — LGTM (non-binding; cannot self-approve)

**Verdict:** Solid Phase 2 wire. Cascade ordering matches spec intent, fall-through is safe, test coverage is comprehensive.

### What I checked

**Cascade correctness vs. spec line 34**
Spec says: "when canDelegate=true, load team-leader.md soul AND team-leader/role-boundaries.md FIRST, then append developer (or whatever original role) as Capability Specialty."

`soul.module.ts` resolution chain after this PR:
1. Per-member soul (highest precedence — explicit > default)
2. **NEW: TL overlay** (canDelegate=true → team-leader.md + Capability Specialty)
3. Role default soul
4. Archetype
5. Hardcoded fallback

`role-boundary.module.ts` resolution chain:
1. Orchestrator-specific
2. Self-watch
3. **NEW: TL overlay** (canDelegate=true → role-boundaries.md)
4. loadRoleFragment (existing)
5. buildTeamLeadBoundary inline (legacy)
6. executor/owner inline builders

Both insertions correctly preempt the legacy IC-role cascade for `canDelegate=true` agents — this is the core fix that flips Sam from "developer with extra responsibilities" to "TL with developer specialty". ✓

**Recursion + override guards**
- `resolveSpecialtySoul()` early-returns null when `role === 'team-leader'` — prevents infinite recursion when the TL archetype role is itself wired with canDelegate. Tested. ✓
- Per-member soul wins over the TL overlay — explicit override semantics. Tested. ✓ (See note below.)

**Fall-through safety**
The PR description warns that Phase 2 ships before Phase 1 markdown is guaranteed everywhere. Both modules treat a missing asset as a no-op fall-through to the legacy cascade rather than failing. The "falls through to legacy cascade when team-leader.md is missing" tests pin this invariant on both modules. ✓ Critical for deployment ordering safety.

**Negative-path guards**
The "does NOT load TL overlay when canDelegate is false" tests use a strict-equality check (`config.canDelegate === true` in source) — so undefined / null / falsy values can't accidentally promote. Mock fs returns the TL fixture if the path is read; the assertion proves the path is never read. Defensive guard against a future `Boolean(config.canDelegate)` style refactor. ✓

**Capability Specialty layering**
`formatTLOverlay()` produces:
- "## Your Soul" + "_Source: team-leader overlay (canDelegate=true)_" + TL content
- "## Capability Specialty" + "_IC background as {role} — apply when self-implementing per the Self-Implementation Exception Rule._" + role specialty

The connecting prose ("apply when self-implementing per the Self-Implementation Exception Rule") explicitly bridges the TL primary identity to the IC specialty — gives the model a frame for *when* to draw on the IC content rather than treating it as a co-equal soul. This is the right framing. ✓

**Member-soul-wins precedence audit**
Spot-checked `~/.crewly/teams/` — found **zero** `soul.md` files anywhere in member dirs (verified via `find ~/.crewly/teams -name soul.md`). So today no member-specific soul will mask the TL overlay when Phase 2.5 flips Sam's role. ✓ Worth re-running this check immediately before the Sam flip in Phase 2.5; documenting now so it doesn't get missed.

### Code quality observations

- **Minor DRY**: `readFileIfExists` is now defined locally in both `role-boundary.module.ts` and `soul.module.ts` (the new one explicitly says "mirrors the SoulModule pattern"). Two definitions of the same trivial helper is acceptable, but if you touch this area again, consider lifting it to `prompt-module.interface.ts` so the contract is shared. Not blocking.
- **JSDoc clarity**: The updated `resolveSoul()` chain comment in soul.module.ts is well-explained — including the deliberate "step 0" framing for the TL overlay and the CSS-specificity analogy in the inline comment. Future readers will understand the precedence.
- **Test fixture pattern**: `mockedFs.readFileSync.mockImplementation((p) => p.endsWith(path.join(...)) ? fixture : throw)` is the right shape — uses `path.join` for cross-platform safety, throws ENOENT-equivalent for unmatched paths. Consistent across all 11 new tests.

### Non-blocking observations

1. **Per-member soul precedence**: Defensible choice (explicit > default), but creates a footgun for the future: any time someone ships a member-specific `soul.md` for a TL agent, that agent silently loses TL framing. Verified zero such files exist today. Suggest:
   - (a) Phase 2.5 pre-flip checklist explicitly includes the `find ~/.crewly/teams -name soul.md` audit, OR
   - (b) consider logging a debug note from `SoulModule` when a per-member soul masks an available TL overlay (so the masking is visible in agent debug logs). Optional.

2. **Legacy `buildTeamLeadBoundary` cleanup**: Once Phase 1 markdown is universally deployed, the inline `buildTeamLeadBoundary` in role-boundary.module.ts becomes dead code (the new TL overlay branch always wins for canDelegate=true). Spec calls for a "net-zero offset" cleanup — track as a Phase 2-followup so the codebase doesn't accumulate two sources of TL boundary truth long-term.

3. **DUMMY tl-test-member smoke**: PR description mentions dummy member assembly verification. I'd recommend running it AFTER Phase 1 markdown is on disk in every test environment (which is true here since Phase 1 commits are stacked under this branch) but BEFORE merging — confirms the assembled prompt ordering matches expectations end-to-end (TL primary → Capability Specialty → role-boundaries.md → org context), not just unit-mock-level.

### Recommendation
Merge after #414 lands. The wire is correct, the tests pin the right invariants, the fall-through is safe, and the override semantics are explicit and tested. The Sam role flip (Phase 2.5) is correctly scoped out — keep it that way; one runtime change at a time.

— Architect (crewly-architect-arch-57d26b22)

PR #418, opened 2026-05-04
docs(orc-prompt): add Periodic Progress Check-In section (user-request cadence)

stevehuang0115:
## Summary
Adds a new section to `config/roles/orchestrator/prompt.md` between **Silent by Default** and **The Owner** that mandates periodic 15-minute check-ins during user requests longer than ~10 minutes.

Pure markdown change — `+36 / -0`, additive only, no edits to existing sections.

## Why
Silent Mode is correct as the default **outside** a user request, but silence **during** an open request reads as "stuck" or "forgot" — the opposite of what the owner wants. There was no built-in cadence rule, so behaviour was inconsistent across orc sessions.

## What changes
- **Default cadence**: every 15 min via the real `schedule-check` skill with `recurring: true`. Capture the schedule ID.
- **On delivery (or owner cancel)**: call `cancel-schedule` to stop the recurring fire — explicit guard against stale schedules.
- **Bound recurring checks** with `maxOccurrences` for short requests (matches `schedule-check` Best Practices in its own `SKILL.md`).
- **Format**: lead with progress, ETA only if changed, blockers only when real. Includes good + bad examples.
- **Override**: owner can mute (`only tell me when done`), change cadence (`every 5 min`), or set a future preference.

## Acceptance criteria (from spec)
- [x] Inserted at a coherent position (after `Silent by Default`, before `The Owner`) — does not break existing structure.
- [x] References the real skills `schedule-check` + `cancel-schedule` (NOT `cancel-followup`, which does not exist in this repo).
- [x] Default 15 min, with explicit override clause.
- [x] Complementary to **Owner-Facing Communication SOP** (P0-6) — this section governs **WHEN** to check in; that SOP governs **HOW** to write the message. Cross-references real existing headings (`Jargon Hygiene`, `Chat/Slack rules`) since the SOP is in a separate doc.

## Tier
Fast — pure markdown, no code paths touched, no test changes required.

## Related
- P0-6 Owner-Facing Communication SOP (complementary, not blocking)

### DIFF
--- a/config/roles/orchestrator/prompt.md
+++ b/config/roles/orchestrator/prompt.md
@@ -50,6 +50,42 @@ The owner hired you to deliver results, not to narrate progress or ask permissio
 
 ---
 
+## Periodic Progress Check-In (User Requests)
+
+Silent Mode is the default **outside** a user request. **Inside** a user request that is going to take more than ~10 minutes to deliver, you MUST keep the owner in the loop with periodic check-ins. Silence during an open request reads as "stuck" or "forgot" — the opposite of the intended behaviour.
+
+**Default cadence:** every 15 minutes until the request is delivered or cancelled.
+
+**Mechanism (use the real scheduling skills, do not roll your own timer):**
+1. Immediately after acknowledging a long-running request, schedule a recurring check by calling **`schedule-check`** with `{"minutes":15,"message":"<request summary>","recurring":true}`. Capture the returned schedule ID.
+2. At each tick (and whenever the schedule fires you back), reply to the owner on the original channel with a short status message — see format below.
+3. When the deliverable is shipped (or the owner cancels), call **`cancel-schedule`** with the captured schedule ID. **Do not leave a recurring check live after the request closes** — that turns into stale chatter.
+
+**Bound the schedule:** for a request you expect to finish within an hour or two, prefer `maxOccurrences` (e.g. 6 ticks for 90 min) over an open-ended recurring check. Stale schedules are a known footgun (see `schedule-check` Best Practices).
+
+**Format of each check-in (1–2 sentences, lead with progress):**
+- Current state — phase / step / PR draft URL.
+- ETA — only mention if it changed since the last check-in.
+- Blockers — call out the specific decision or input you need; otherwise omit.
+- Follow the Jargon Hygiene rules below — no internal IDs, session names, or skill names in the message itself.
+
+**Examples (good):**
+- "Phase 2 of 3 done — backend wire merged, frontend hookup in review. ETA still on track for tonight."
+- "Draft PR up — https://github.com/.../pull/417. Waiting on your call: target main now or stack on #414?"
+
+**Examples (bad — don't do this):**
+- "Still working." (no progress, no ETA — useless)
+- "Sam is in_progress on Phase 2 builder wire, Leo is idle, Mia is reviewing." (internal team chatter — owner doesn't care)
+
+**Override (the owner controls this):**
+- Owner says "don't check in" / "只在做完时告诉我" / "only tell me when done" / "stop the updates" → call `cancel-schedule` immediately and revert to silent-until-done for this request.
+- Owner says "check in every 5 min" / "更频繁一点" → cancel the existing schedule, re-schedule with the new interval.
+- Owner says "next time tell me less / more" → adjust the default cadence for *future* requests in this conversation.
+
+**Rule of thumb:** Silent by default **outside** a user-request flow; periodic check-in **inside** one. This section governs **when** to ping; the *how* (jargon, tone, formatting) is governed by **Jargon Hygiene** and the **Chat/Slack rules** elsewhere in this prompt.
+
+---
+
 ## The Owner (your primary audience)
 
 You are talking to a **small-business owner**, not a developer or a team-mate on your floor. Default assumption: they know the business outcomes, not the internal machinery.


stevehuang0115 (commented):

Light review of #418. Pure markdown, additive, +36/-0, single file. Net: ship it.

---

### What this PR does

Adds a `## Periodic Progress Check-In (User Requests)` section to the orchestrator prompt between **Silent by Default** (line 29) and **The Owner** (line 53). Establishes a 15-minute default check-in cadence for long-running user requests, with explicit mechanism (use the real `schedule-check` + `cancel-schedule` skills, not a hand-rolled timer), format, and owner-override clause.

Closes a real behavioral gap: Silent Mode is correct OUTSIDE a request, but silence DURING an open request reads as "stuck" or "forgot" — the opposite of intended behavior. There was no codified cadence rule before this PR, so different orc sessions handled it inconsistently.

---

### Reality-check on every skill / heading the new section references

| Reference | Status | Notes |
|---|---|---|
| `schedule-check` skill | ✅ exists | `config/skills/orchestrator/schedule-check/execute.sh` + SKILL.md confirm the exact invocation shape used in the PR (`{minutes, message, recurring, maxOccurrences}`) |
| `cancel-schedule` skill | ✅ exists | `config/skills/orchestrator/cancel-schedule/` |
| `cancel-followup` (NOT used) | ✅ correctly avoided | PR description explicitly calls out: *"References the real skills schedule-check + cancel-schedule (NOT cancel-followup, which does not exist in this repo)."* That's a `agent/core` skill, not an orchestrator skill — getting this right matters because the orc would otherwise try to invoke a non-existent path. ✅ |
| `## Silent by Default` cross-reference | ✅ exists | line 29 — placement of new section directly after this is coherent |
| `## Jargon Hygiene` cross-reference | ✅ exists | line 71 — used to govern the *how* of each check-in message (no internal IDs, session names, skill names) |
| "Chat/Slack rules" reference | ⚠️ soft reference | Not a literal heading; refers to the broader `## Your Markdown Content` / `### How to Respond to Chat Messages` / `### The reply-slack Skill (Slack)` family at lines 277-379. Acceptable as a soft reference but not a strict link. Minor — see observation 1 below. |
| `maxOccurrences` parameter | ✅ exists | confirmed in `schedule-check/SKILL.md` Parameters table + Best Practices: *"Use `maxOccurrences` for time-bounded tasks to prevent stale schedules from running indefinitely."* PR's `e.g. 6 ticks for 90 min` example is exactly that pattern. |

Every concrete reference checks out against the real skill contracts and the real prompt structure.

---

### Content quality

- **Clear trigger**: "more than ~10 minutes to deliver" — soft heuristic but actionable. The default+override pattern means inconsistency at the trigger boundary is bounded by the owner's veto.
- **Default cadence (15 min)** — sensible middle ground between noise and silence.
- **Mechanism ties to real skills** with correct param shape, including the `maxOccurrences` best-practice for short requests.
- **"Don't roll your own timer"** clause is the right defensive directive — prevents the orc from approximating cadence with `await new Promise(setTimeout(..., 15*60*1000))` shenanigans.
- **Format guidance** (lead with progress, ETA only on change, blockers only when real) maps well to the existing Jargon Hygiene contract.
- **Good/bad examples** are concrete and demonstrative — particularly the "Sam is in_progress on Phase 2 builder wire, Leo is idle, Mia is reviewing" anti-example, which is exactly the kind of internal team chatter the owner doesn't care about. ✅
- **Override clause** is comprehensive (mute, change cadence, set future preference) and bilingual (English + Chinese trigger phrases) — matches the bilingual owner-audience pattern visible elsewhere in the prompt.
- **Closing rule of thumb** ("Silent by default OUTSIDE a user-request flow; periodic check-in INSIDE one") cleanly resolves the apparent tension with the existing Silent by Default section.

---

### Observations (non-blocking, none worth holding up merge)

1. **"Chat/Slack rules" is a soft reference**, not a literal heading. Could either change to "the **Your Markdown Content** and **reply-slack** sections below" for precision, or leave as-is for readability. Tradeoff is taste; current phrasing works.
2. **Schedule ID storage isn't specified.** "Capture the returned schedule ID" — but where? In conversation context? In a per-session state file? An orc session that gets compacted could lose the ID and then be unable to call `cancel-schedule` cleanly. Worth a P2 follow-up: pair `schedule-check` with a state-file write or recommend the orc note the ID inline in the acknowledgment message so it survives compaction.
3. **Bilingual examples asymmetry.** Override-clause trigger phrases include Chinese (`只在做完时告诉我`, `更频繁一点`), but the good/bad check-in examples are English-only. Owners writing in Chinese might benefit from one bilingual example. Minor.
4. **Threading model unspecified.** "Reply to the owner on the original channel" — but if the owner asked on Slack, does the check-in thread under the original message or post fresh? Probably handled by the existing `reply-slack` skill's defaults, but a one-line clarification ("thread under the original request when on Slack") would tighten the contract. Non-blocker since reply-slack semantics likely cover it.
5. **Trigger threshold (~10 min) is operator-judged.** Some orc sessions will over-call check-ins, some under-call. Could be tightened with a more concrete heuristic (e.g. "if you delegated to a team and don't expect the first deliverable for >10 min"), but that's gilding — the override clause means the owner gets the final word on cadence anyway.

---

### Recommendation

**Merge as-is.** Tight markdown, additive only, every skill + heading reference verified, content fills a real behavioral gap. Owner-override clause keeps the cadence under owner control, which is the right escape hatch.

**Suggested follow-ups (none blocking, all P2):**
1. Document where the orc should persist the captured schedule ID so it survives session compaction.
2. Add one bilingual check-in example to match the bilingual override-trigger phrases.
3. Optionally tighten the "Chat/Slack rules" cross-reference to point at the literal section names (`Your Markdown Content` / `reply-slack`).

PR #400, opened 2026-05-01
feat(crewly-agent): add DeepSeek provider via OpenAI-compatible API

stevehuang0115:
## Summary
- Adds `'deepseek'` to the `ModelProvider` union in `crewly-agent/types.ts` and wires it through `ModelManager` so agents can be configured with `provider: 'deepseek', modelId: 'deepseek-chat'`.
- Implementation reuses the already-installed `@ai-sdk/openai` package via `createOpenAI({ baseURL: DEEPSEEK_BASE_URL, apiKey }).chat(modelId)` — DeepSeek's API is OpenAI-compatible, so **no new npm dependency**.
- API key is read from `process.env.DEEPSEEK_API_KEY` directly. Follow-up tracked in **#401**.

## Why
Steve requested DeepSeek support so the `crewly-agent` runtime can be evaluated against deepseek-chat / deepseek-reasoner alongside Anthropic / OpenAI / Google / Ollama. ETA per Steve: 1–2 days through normal cadence; this PR follows the standard merge → next 1.6.x release path (no hotfix, no rsync).

## Files touched (4, ~55 LOC, +`DEEPSEEK_BASE_URL` const)
- `backend/src/services/agent/crewly-agent/types.ts` — extend `ModelProvider` + `MODEL_PROVIDERS`
- `backend/src/services/agent/crewly-agent/types.test.ts` — `MODEL_PROVIDERS` length now 5
- `backend/src/services/agent/crewly-agent/model-manager.ts` — `'deepseek'` switch case routes via `.chat(modelId)` (NOT bare call — DeepSeek does not implement `/responses`); module-local `DEEPSEEK_BASE_URL` const; `getAvailableProviders` includes `deepseek`; `ensureApiKeyInEnv` short-circuits for deepseek
- `backend/src/services/agent/crewly-agent/model-manager.test.ts` — positive case for DeepSeek model creation, regression guard `provider === 'openai.chat'`, env-var detection in `getAvailableProviders`. Unknown-provider negative case (`'azure'`) preserved.

## Critical fix from review (M1)
Per Sam's review: the bare `deepseekProvider(modelId)` form on `@ai-sdk/openai >= 3.x` routes to `/responses`, which DeepSeek does **not** implement (only `/chat/completions`). Updated to `deepseekProvider.chat(modelId)`. Test now asserts `(model as any).provider === 'openai.chat'` so a future refactor cannot silently re-introduce the bug.

## ConfigService / env-var deployment note (for orc)
[...]

PR #419, opened 2026-05-04
feat(crewly-agent): DeepSeek Phase 1 — 5 BLOCKER fixes for daily ops

stevehuang0115:
## Summary

Phase 1 of the DeepSeek gap-list — ships the 5 BLOCKERs needed for Crewly Agent runtime to run `deepseek-chat` and `deepseek-reasoner` for daily ops.

Spec: `.crewly/specs/2026-05-03-crewly-agent-deepseek-gap-list.md`

| # | Fix | Files | Tests |
|---|---|---|---|
| **B1** | DeepSeek wired through settings service (skill→runtime→global→env), provider parity in `API_KEY_PROVIDERS` / `API_KEY_ENV_VARS` / `isValidApiKeyProvider` | `settings.types.ts`, `model-manager.ts` | +8 settings.types, +2 model-manager |
| **B2** | `SUPPORTED_MODELS` catalog includes `deepseek-chat` + `deepseek-reasoner` w/ 64k context | `types.ts` | +3 types |
| **B3** | `resolveContextWindow` returns 64_000 for both DeepSeek models | `types.ts` | +3 types |
| **N5** | `resolveMaxOutputTokens` floor — honors `model.maxTokens` without undershooting model floor | `types.ts` | +3 types |
| **B4** | **Regression test only.** INVALIDATED via live smoke (3/3 deterministic, 1.4–1.5s, all picked `reply_slack`, 0 fallback — spec L126-140). Pins passthrough: agent-runner does NOT inject `toolChoice`; relies on AI SDK `'auto'` default. | (test only) | +3 agent-runner |

### Why B4 has no product code

```
runs: [
  { run: 1, ms: 1528, toolName: 'reply_slack', stepCount: 1, err: null },
  { run: 2, ms: 1397, toolName: 'reply_slack', stepCount: 1, err: null },
  { run: 3, ms: 1411, toolName: 'reply_slack', stepCount: 1, err: null },
]
all_picked_reply_slack: true
```

Adding a `toolChoice: 'required'` retry layer was unnecessary risk against deterministic behavior. The 3 regression tests fail loudly if anything in the future starts injecting `toolChoice` — forcing a deliberate re-eval against this evidence.

### Pre-existing fixes (unblocking, not scope creep)

`backend/src/types/settings.types.test.ts` had 3 `GeneralSettings` fixtures missing `enableAuditor: false` — pre-existing on `main`, blocked the file from compiling. Added the field so my B1 tests could run in the same file. Verified pre-existing via stash + rerun on `398ec464`.

## Test plan

- [x] `npx jest --testPathPattern='types/settings'` → **82/82 ✅**
- [x] `npx jest --testPathPattern='crewly-agent/types'` → **48/48 ✅**
- [x] `npx jest --testPathPattern='model-manager'` → **17/17 ✅**
- [x] `npx jest -t 'B4'` → **3/3 ✅**
- [x] `npx tsc --noEmit` on `backend/` → clean
- [x] Full sweep `crewly-agent + types/settings`: 820 pass, 5 fail — all 5 fails are pre-existing on `main` in `tool-registry.test.ts` (`senderSessionName` field, not touched by this PR; verified via stash on `398ec464`)
- [ ] Reviewer: please run live smoke on `deepseek-chat` against an actual orchestrator to confirm Settings UI input → settings service → ensureApiKeyInEnv → DeepSeek model creation chain end-to-end

## Out of scope (Phase 2 candidates)

- Frontend Settings UI input box for `DEEPSEEK_API_KEY` (~30min, separate PR)
- Streaming path tool_choice passthrough hardening (covered in B4 regression but `streamText` path not exercised — only `generateText`)
- Pre-existing tool-registry `senderSessionName` failures

### DIFF
--- a/backend/src/services/agent/crewly-agent/agent-runner.service.test.ts
+++ b/backend/src/services/agent/crewly-agent/agent-runner.service.test.ts
@@ -2168,3 +2168,109 @@ describe('AgentRunnerService.checkMissingDeliverables (P0)', () => {
     expect(missing).toEqual([]);
   });
 });
+
+/**
+ * B4 — DeepSeek tool_choice passthrough regression test.
+ *
+ * Spec: /Users/yellowsunhy/Desktop/projects/crewly-projects/crewly/.crewly/specs/2026-05-03-crewly-agent-deepseek-gap-list.md
+ *
+ * Background — B4 was originally listed as a 🔴 BLOCKER ("tool_choice 不确定")
+ * estimated at 2h. Live smoke testing on 2026-05-03 INVALIDATED it:
+ *
+ *     ✅ Live test 3/3 runs, toolChoice: { type: 'tool', toolName: 'reply_slack' }
+ *        on deepseek-chat V3 = completely deterministic. Each run returned the
+ *        correct tool_call within 1.4-1.5s, 0 fallback to text, 0 multi-tool drift.
+ *
+ *     runs: [
+ *       { run: 1, ms: 1528, toolName: 'reply_slack', stepCount: 1, err: null },
+ *       { run: 2, ms: 1397, toolName: 'reply_slack', stepCount: 1, err: null },
+ *       { run: 3, ms: 1411, toolName: 'reply_slack', stepCount: 1, err: null },
+ *     ]
+ *     all_picked_reply_slack: true
+ *
+ * Decision (TL): no product-code change for B4 — DO NOT add a retry-with-tool-choice
+ * fallback layer. This regression test pins the current "passthrough" behavior:
+ * agent-runner.service.ts must NOT inject a hardcoded `toolChoice` into the
+ * generateText / streamText call. The default ('auto') lets the model decide,
+ * and the live smoke confirms DeepSeek V3 is reliable under that default.
+ *
+ * If a future change adds `toolChoice: 'required'` or similar to either
+ * the streamText or generateText path, these tests fail — forcing a re-evaluation
+ * against the smoke evidence above.
+ */
+describe('B4 — DeepSeek tool_choice passthrough regression', () => {
+  let runner: AgentRunnerService;
+  let mockGenerateText: jest.Mock<any>;
+
+  const baseConfig: CrewlyAgentConfig = {
+    model: { provider: 'deepseek', modelId: 'deepseek-chat', temperature: 0.3, maxTokens: 8192 },
+    maxSteps: 10,
+    sessionName: 'b4-regression-session',
+    apiBaseUrl: 'http://localhost:8787',
+    systemPrompt: 'You are a deepseek agent.',
+    maxHistoryMessages: 20,
+    compactionThreshold: 0.8,
+  };
+
+  beforeEach(async () => {
+    jest.clearAllMocks();
+
+    mockGenerateText = jest.fn<any>().mockResolvedValue({
+      text: 'noop',
+      steps: [{ toolCalls: [], toolResults: [] }],
+      usage: { inputTokens: 10, outputTokens: 5 },
+      finishReason: 'stop',
+    });
+
+    const mockModelManager = {
+      getModel: jest.fn<any>().mockResolvedValue({ provider: 'deepseek', modelId: 'deepseek-chat' }),
+      getAvailableProviders: jest.fn<any>(),
+      clearCache: jest.fn<any>(),
+    } as any;
+
+    const mockApiClient = {
+      get: jest.fn<any>(),
+      post: jest.fn<any>(),
+      delete: jest.fn<any>(),
+    } as any;
+
+    runner = new AgentRunnerService(baseConfig, mockModelManager, mockApiClient);
+    runner._generateTextFn = mockGenerateText;
+    await runner.initialize();
+  });
+
+  it('should NOT pass a hardcoded toolChoice to generateText (let SDK default apply)', async () => {
+    await runner.run('say hi');
+
+    expect(mockGenerateText).toHaveBeenCalled();
+    const callArgs = mockGenerateText.mock.calls[0]?.[0] as Record<string, unknown>;
+    expect(callArgs).toBeDefined();
+
+    // The contract: agent-runner relies on the AI SDK's default toolChoice='auto'.
+    // DeepSeek V3 was validated as deterministic under this default (3/3 smoke runs).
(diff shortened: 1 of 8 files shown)


stevehuang0115 (commented):

Architectural review of #419 against the 4 focus points + a follow-up note. Net: ship it.

---

### 1. crewly-agent runtime integration cleanliness — ✅ clean

`types.ts`, `model-manager.ts`, `agent-runner.service.ts` all hold up.

- **`types.ts`** — `MODEL_CONTEXT_WINDOWS` and `MODEL_OUTPUT_TOKEN_FLOORS` are introduced as flat lookup tables keyed by `modelId`, which is the right shape. `SUPPORTED_MODELS` adds the two DeepSeek IDs in the same `provider/modelId` form as the rest. No special-casing of DeepSeek in the type layer — it is treated as a peer of openai/anthropic/google. Good.
- **`model-manager.ts`** — Removing the `if (provider === 'deepseek') return;` short-circuit in `ensureApiKeyInEnv` and adding the `case 'deepseek': process.env.DEEPSEEK_API_KEY = key; break;` writeback is exactly the symmetric move. The `Exclude<ModelProvider, 'ollama' | 'deepseek'>` → `Exclude<ModelProvider, 'ollama'>` widening on `providerToApiKeyProvider` is the correct downstream consequence. `getAvailableProviders` correctly switches from `!!process.env.DEEPSEEK_API_KEY` to the settings-service fetch, so the UI now sees a coherent availability signal.
- **`agent-runner.service.ts`** — All 4 sites that previously read `this.config.model.maxTokens` directly (streamText 791-792, generateText 989-990, follow-up 1196-1197, compaction 1265-1266) now go through `resolveMaxOutputTokens(this.config.model)`. Comprehensive — no escape hatch left.

No layering violations, no tight coupling to DeepSeek specifics in the runner — the floor logic stays inside the helper.

---

### 2. N5 `resolveMaxOutputTokens` — ✅ design is right

```ts
export function resolveMaxOutputTokens(config: ModelConfig): number {
  const requested = config.maxTokens ?? CREWLY_AGENT_DEFAULTS.DEFAULT_MODEL.maxTokens ?? 0;
  const floor = MODEL_OUTPUT_TOKEN_FLOORS[config.modelId] ?? 0;
  return Math.max(requested, floor);
}
```

Two lines, three correct decisions:
- **`?? 0` fallback chain** handles the truly-unset case without throwing, and lets `Math.max` do the right thing.
- **`Math.max(requested, floor)`** rather than "always force floor" — preserves user intent when they explicitly request more, only clamps up when they're below the safe minimum. This is the right semantics for R1's `reasoning_tokens` + `completion_tokens` mixing-into-`max_tokens` bug.
- **Floor of 1024 for `deepseek-reasoner`** — empirically grounded; R1 burns several hundred tokens on chain-of-thought before emitting a single visible character, so 512 is borderline and 256 silently empties output. 1024 leaves headroom without being wasteful for short answers.

Lookup-table form (`MODEL_OUTPUT_TOKEN_FLOORS`) scales cleanly when other reasoning models land. No reason to refactor preemptively.

---

### 3. B4 regression test coverage — ✅ correctly pins the invariant

The three tests (`agent-runner.service.test.ts`) inject `runner._generateTextFn = mockGenerateText` (test seam confirmed at `agent-runner.service.ts:183`) and assert the right thing: `expect(callArgs).not.toHaveProperty('toolChoice')` on first call, every multi-turn call, and on a tools-bearing call. This is the correct way to pin "AI SDK default behavior is what we ship" — testing absence of an override, not presence of a specific value.

**Coverage gap (acknowledged in PR description, not a blocker):** the `streamText` path at `agent-runner.service.ts:791` is not exercised by these tests — only the `generateText` path is. Since the runner uses both paths in different conditions, ideally the same invariant is pinned for streamText. PR description correctly scopes this as Phase 2 work; it's tracked.

Test-only change, no product code → low risk.

---

### 4. Incremental relationship to PR #400 — ✅ clean telescope

Verified via `gh pr view 400`. PR #400 added:
- `case 'deepseek'` factory in `getProviderFunction` (with the correct `.chat(modelId)` routing — the `/chat/completions`-only DeepSeek constraint).
- `SUPPORTED_MODELS` entries.
- Env-var-only access (`DEEPSEEK_API_KEY` directly).

PR #419 layers on top:
- Settings-service wiring for parity with other cloud providers (B1).
- The R1 max-tokens floor (N5).
- The toolChoice regression test (B4 — independent of DeepSeek but bundled because the same dev hit it).
- Type/test surface for `ApiKeyProvider` including `deepseek`.

No overlap, no rework of #400's code paths. The two PRs telescope properly: #400 = "DeepSeek works", #419 = "DeepSeek is a first-class citizen."

---

### Non-blocking follow-up worth filing

**Provider-cache staleness for DeepSeek key rotation.** In `model-manager.ts:104-117`, the DeepSeek factory captures `process.env.DEEPSEEK_API_KEY` at `createOpenAI()` construction time (line 113), then the resulting closure is cached in `this.providerCache` (line 131). If a user rotates their DeepSeek key via the Settings UI mid-runtime, `ensureApiKeyInEnv` correctly writes the new value to `process.env`, but the cached factory still calls `deepseekProvider.chat(modelId)` against the stale `deepseekProvider` instance built with the old key.

This is **unique to DeepSeek** among current providers. `@ai-sdk/anthropic`, `@ai-sdk/openai`, `@ai-sdk/google` read env vars at request time, so they're naturally key-rotation-safe. DeepSeek is forced through `createOpenAI({ apiKey: ... })` because it needs a custom `baseURL`, which captures the key eagerly.

It's **pre-existing from #400**, not introduced here — but #419 makes it more user-visible because the Settings UI is now the primary config path for DeepSeek (rather than `.env` + restart). On `.env`-only flow there was no rotation surface; now there is.

Three reasonable fixes for a follow-up:
1. Invalidate the DeepSeek cache entry inside `ensureApiKeyInEnv` when the key changes.
2. Pass `apiKey: () => process.env.DEEPSEEK_API_KEY` (function form) to `createOpenAI`, if the SDK supports it — defers the env read.
3. Document that DeepSeek key rotation requires a runtime restart and surface that in the Settings UI.

I'd lean (2) if the SDK supports function form, else (1). Not a blocker for #419.

---

**Recommendation:** merge #419 as-is. File the cache-staleness follow-up as a separate issue and address in Phase 2 alongside the streamText `toolChoice` test gap.

PR #420, opened 2026-05-04
fix(persistence-p0): integrity-guarded writes + rotating snapshots + boot invariant

stevehuang0115:
## Summary

P0 fix for the 2026-05-03 silent-state-wipe incident, per `.crewly/specs/2026-05-03-state-persistence-fix-spec.md`. Three independent layers so any single failure cannot collapse a healthy state to `[]`:

- **A1 — Integrity-guarded atomic writes.** New `atomicWriteJsonWithGuard<T>(path, next, opts)` wrapper (`backend/src/utils/integrity-guarded-write.utils.ts`) that compares prev count vs next count before swapping the file. Refuses (a) collapse-to-empty (`prev > 0 && next === 0`) and (b) decrease beyond a configurable threshold (default 50%). Surfaces a typed `IntegrityViolationError`. Wired into the 5 aggregate-write sites:
  - `StorageService.saveProject`
  - `StorageService.saveScheduledMessage`
  - `StorageService.saveRecurringCheck`
  - `StorageService.saveOneTimeCheck`
  - `TeamsBackupService.updateBackup` (custom countOf for `{timestamp, teams[]}` shape)

  Delete paths intentionally NOT wired — those are explicit operator-driven removals.

- **A2 — Rotating snapshot history (ring buffer, N=30).** `TeamsBackupService` now writes every backup to `teams-backup-history/teams-backup-NNN.json` plus an `index.json {head, written, slots}` BEFORE the live single-file backup. Slot file first, then index — a crash leaves at most an unreferenced slot file, never a torn index. New restore APIs: `getBackupHistory()` (newest-first), `readSlot(n)`, `restoreFromSlot(n)`. History writes are best-effort and never block the live backup path.

- **C1 — Boot-time state invariant check.** New `StorageService.verifyStateInvariantOnBoot()` runs in `backend/src/index.ts` BEFORE `startHttpServer()`. If live teams are empty but the backup is populated, it throws `StateInvariantViolation` and the backend refuses to serve traffic — preventing the UI from being shown an empty state that the user might "fix" by recreating teams against the wiped data, destroying recovery options. Operator override: `CREWLY_FORCE_EMPTY_BOOT=1` for legitimate fresh-installs / explicit resets. Best-effort failure mode: if the check itself errors, it logs a warn and proceeds — the invariant must never become a boot-time hard dependency.

## Test plan

- [x] A1 wrapper unit tests — 25/25
- [x] A2 ring-buffer tests + A1 wire-in regression — 25/25
- [x] StorageService aggregate-write tests — 20/20
- [x] C1 unit tests (error shape, env-var parsing) — 9/9
- [x] C1 integration tests (real fs + real TeamsBackupService) — 7/7
- [x] **Full P0 unit/integration suite: 76/76 passing** across 6 test files
- [x] **e2e artificial-nuke test (acceptance gate) — 3/3 passing** (`backend/src/services/core/persistence-recovery.e2e.test.ts`, commit `f91cd608`):
  - full chain: healthy state → nuke `teams/` → fresh services → boot refused with `StateInvariantViolation` → `restoreFromSlot(0)` → re-seed via `saveTeam` → boot clean
  - `CREWLY_FORCE_EMPTY_BOOT=1` operator override path
  - A1 guard rejects in-flight collapse-to-empty `updateBackup`
- [ ] Full repo `npm run build && npm test` (final verification)
- [ ] Manual smoke: create team, observe ring-buffer slots populating in `~/.crewly/teams-backup-history/`

## Commits (5)

| SHA | Subject |
|---|---|
| `851090b7` | A1 — integrity-guarded atomic write wrapper |
| `dd27c95b` | A2 — rotating snapshot history (ring buffer) |
| `505a3bfe` | A1 wire-ins — 5 aggregate-write sites |
| `50932fe8` | C1 — boot-time state invariant check |
| `f91cd608` | e2e artificial-nuke recovery test (acceptance gate) |

## Recovery history note

The original A1 commit (`6f679e7e`) was lost during a local branch reset; the wrapper was re-created from spec on top of `main` as `851090b7` (identical subject, different SHA). All four feature commits visible in `git log main..HEAD` were brought back via cherry-pick / rewrite from the lost branch and verified test-clean before push. Reflog still has the original SHAs for audit. No work was lost; just SHAs.

## Spec reference

`.crewly/specs/2026-05-03-state-persistence-fix-spec.md` — sections A1, A2, C1, acceptance gate.

### DIFF
--- a/backend/src/controllers/teams-backup/teams-backup.controller.test.ts
+++ b/backend/src/controllers/teams-backup/teams-backup.controller.test.ts
@@ -7,10 +7,16 @@
 import { describe, it, expect, beforeEach, jest } from '@jest/globals';
 import type { Request, Response, NextFunction } from 'express';
 import type { Team } from '../../types/index.js';
+import type {
+  TeamsBackup,
+  TeamsBackupHistoryEntry,
+} from '../../services/core/teams-backup.service.js';
 
 // Mock data holders
 const mockTeams: { value: Team[] } = { value: [] };
-const mockBackup: { value: { timestamp: string; teams: Team[] } | null } = { value: null };
+const mockBackup: { value: TeamsBackup | null } = { value: null };
+const mockHistory: { value: TeamsBackupHistoryEntry[] } = { value: [] };
+const mockSlots: { value: Record<number, TeamsBackup> } = { value: {} };
 
 jest.mock('../../services/core/storage.service', () => ({
   StorageService: {
@@ -31,6 +37,8 @@ jest.mock('../../services/core/teams-backup.service', () => ({
         backupTimestamp: mockBackup.value?.timestamp || null,
       }),
       readBackup: async () => mockBackup.value,
+      getBackupHistory: async () => mockHistory.value,
+      restoreFromSlot: async (slot: number) => mockSlots.value[slot] ?? null,
     }),
   },
 }));
@@ -48,7 +56,12 @@ jest.mock('../../services/core/logger.service.js', () => ({
   },
 }));
 
-import { getBackupStatus, restoreFromBackup } from './teams-backup.controller.js';
+import {
+  getBackupStatus,
+  restoreFromBackup,
+  getBackupHistory,
+  restoreFromSlot,
+} from './teams-backup.controller.js';
 
 /**
  * Helper to create a mock Team.
@@ -85,6 +98,8 @@ describe('TeamsBackupController', () => {
   beforeEach(() => {
     mockTeams.value = [];
     mockBackup.value = null;
+    mockHistory.value = [];
+    mockSlots.value = {};
     jest.clearAllMocks();
   });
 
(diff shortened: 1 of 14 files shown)


stevehuang0115 (commented):

Architectural review of #420 against `.crewly/specs/2026-05-03-state-persistence-fix-spec.md` §6 acceptance criteria. Net: ship it.

I reviewed all 5 commits — A1 wrapper, A2 ring buffer, A1 wire-ins, C1 boot invariant, and the e2e acceptance-gate test that just landed. 79/79 green per Leo. Three independent layers, each independently verifiable, each correctly tested.

---

### A1 — Integrity-guarded atomic write — ✅ design is right

`backend/src/utils/integrity-guarded-write.utils.ts` is well-thought-out:

- **Typed `IntegrityViolationError`** carries `path`, `prevCount`, `nextCount`, and a discriminated `reason: 'collapse-to-empty' | 'decrease-exceeds-threshold'` — callers can branch on cause without string parsing.
- **Two-branch guard** matches spec §4 A1 sketch exactly: collapse-to-empty (line 173) and decrease-exceeds-threshold (line 193). Default `maxDecreasePct: 50`.
- **`allowExplicitDelete: true` opt-out** preserves the legitimate mass-delete path (spec §4 A1 mitigation).
- **Customizable `countOf`** handles non-array shapes — used at the live teams backup site for the `{timestamp, teams[]}` envelope (`teams-backup.service.ts:178`). Correct factoring.
- **Corrupted-prior tolerance** (lines 165-169 + 244-255): if the prior file exists but is unreadable/unparseable, the guard logs a warn and proceeds. **This is the right call** — a corrupted prior file should not block recovery writes. Spec design contract called this out and the implementation honors it.
- **Lazy logger acquisition** (lines 57-63) keeps the module testable when callers mock the logger.
- **Stack trace captured** in the structured log payload — forensic data for post-incident analysis.

Spec acceptance §6 → all four unit cases (decrease-N-to-0, decrease-50%+, allowExplicitDelete bypass, first-ever-write succeeds) covered by the 25 wrapper tests.

---

### A2 — Rotating snapshot ring buffer — ✅ ordering is correct

`teams-backup.service.ts` extends N=1 → N=30 cleanly:

- **Critical ordering is right** (`writeRotatingSnapshot` lines 337 + 347): slot file written FIRST, then index. Spec §4 A2 risk callout: *"interrupted rotation could leave the index pointing at a half-written slot. Atomic-write the slot first, then atomic-write the index."* Implementation matches exactly. A crash between the two writes leaves at most an unreferenced slot file (harmless), never an index pointing at torn data.
- **Best-effort discipline**: history-write failures wrap in try/catch in `updateBackup` (lines 161-167) and never block the live backup path. Spec design intent: *"primary durability of the live store wins."* Honored.
- **Service/controller separation**: `restoreFromSlot()` (line 302) is a **read-only** API — it returns the snapshot, and the controller (`teams-backup.controller.ts:181-194`) handles the re-save loop via `StorageService.saveTeam()`. The service never transitively mutates StorageService, which keeps the dependency graph clean and the test surface manageable.
- **API surface** (`getBackupHistory`, `readSlot`, `restoreFromSlot`, `getHistoryDir`) matches spec §4 A2.
- **Newest-first sort** in `getBackupHistory` (lines 272-276) correctly puts head slot first, then chronological DESC — right shape for an admin restore-picker UI.

Micro-observation (non-blocker): `writeRotatingSnapshot` line 332 has `prevIndex.written === 0 ? 0 : (prevIndex.head + 1) % MAX_HISTORY_SLOTS`. The first branch is redundant — `head = -1` for the empty index produces `(-1 + 1) % 30 = 0` already. Defensive and harmless; flagging only because the dual branches imply asymmetric semantics that don't actually exist.

---

### A1 wire-ins — ✅ pattern matches the failure mode

5 aggregate-write sites guarded:
- `StorageService.saveProject` (`storage.service.ts:886`)
- `StorageService.saveScheduledMessage` (line 1247)
- `StorageService.saveRecurringCheck` (line 1333)
- `StorageService.saveOneTimeCheck` (line 1433) — strictly more defensive than spec listed; same pattern applied uniformly
- `TeamsBackupService.updateBackup` (line 176) — custom countOf for the `{timestamp, teams[]}` envelope shape

**Spec deviation worth naming**: spec §6 acceptance says "**saveTeam** / **deleteTeam** / projects / scheduledMessages / recurringChecks paths swapped." The PR does NOT wire `saveTeam`/`deleteTeam`. This is **architecturally correct**: `saveTeam` writes a single team object to `teams/{id}/config.json` (a per-team file, not an aggregate array), so the decrease-guard semantic doesn't apply. The aggregate "all teams gone" failure mode is caught downstream at `TeamsBackupService.updateBackup` (which IS guarded). The spec wording inferred from the symptom ("teams disappeared") without distinguishing per-team-file vs aggregate-array storage; the PR took the correct path. **Suggest a one-line note in the PR description or a spec update** so future readers don't expect saveTeam to be guarded.

Delete paths intentionally NOT wired — explicit operator-driven removals via plain `atomicWriteFile`. PR description calls this out. Correct.

---

### C1 — Boot-time state invariant — ✅ defensive layering done right

`StorageService.verifyStateInvariantOnBoot` (`storage.service.ts:139`) is exactly what the spec asked for, plus three things the spec only sketched:

1. **Three best-effort try/catch boundaries** (lines 141-150 / 152-164 / 175-185) so the invariant check itself can never wedge boot. Spec §4 C1 risk callout: *"the invariant must never become a boot-time hard dependency itself."* Honored. If `getTeams()` or `getBackupStatus()` throws, the check logs a warn and skips — the existing error paths handle the underlying issue.
2. **Override discoverability**: error message (lines 192-201) includes BOTH restore endpoints (live + per-slot) AND the override env var name. A user hitting the error can self-recover from the message alone. This matters for pilot operators without our triage knowledge.
3. **Override parsing is liberal** (`state-invariant.types.ts:65-70`): accepts `'1' | 'true' | 'yes'`, case-insensitive. Defensive about the user typing `=true` instead of `=1`.

**Boot wire-in placement** (`backend/src/index.ts:1760` → `startHttpServer()` at line 1778): correct ordering. The HTTP server never starts in a wiped state, so the UI never serves an "empty teams" page that the user might "fix" by re-creating teams against the wiped state — this is the recovery-destruction surface the spec was built to close, and it's closed.

**Spec deviation worth naming**: spec §9 Q4 asked: *"After A2 (multi-slot), should `hasMismatch` check against the most-recent-slot count, the max-count across all slots, or both? Recommend max-count across recent N slots."* The PR uses **only the live single-file backup** via `getBackupStatus()`. This is **OK in practice** — A1 now actively prevents the corrupting save from wiping the live backup, so the live-backup-only predicate is sufficient. But the defense is one layer thinner than spec recommended. **Trade-off rationale that I'd accept**: simpler predicate → less false-positive surface; rotating slots are the *recovery* path, not the *detection* path. Defensible scope call. Worth flagging as deliberate, not as oversight.

---

### e2e acceptance gate — ✅ closes the spec ship gate

`persistence-recovery.e2e.test.ts` (244 lines, 3 cases, real fs + real services, no mocks):

1. **5-phase artificial nuke recovery** mirrors spec §7.3 exactly: healthy state → `rm -rf teams/` → fresh service instance → invariant fires with correct counts + remediation hint → `restoreFromSlot(0)` → re-seed → fresh service instance → invariant passes → `getTeams()` returns the original 3 teams by id. This is the spec's named "acceptance gate".
2. **Override regression**: `CREWLY_FORCE_EMPTY_BOOT=1` boots through after nuke. Confirms the legitimate fresh-install / reset path stays open.
3. **In-flight wipe regression** (lines 218-243): explicit `backup.updateBackup([])` against a 3-team prior state. Pre-fix: would have wiped the backup. Post-fix: A1 guard rejects internally, `updateBackup` catch block logs but doesn't re-throw, prior healthy backup survives on disk verbatim. **This is a regression test for the exact 2026-05-03 failure mode** — exactly what we want guarding the regression frontier.

The forceExit flag (LoggerService.startLogFlusher's setInterval handle) is unrelated infra noise that affects any singleton-using test. Acknowledge and move on; not a fix-in-this-PR concern.

---

### Spec acceptance criteria pass/fail

| Criterion (§6 P0) | Status | Notes |
|---|---|---|
| A1 wrapper exists + (a/b/c/d) unit tests | ✅ | 25/25 |
| A1 wired into 5 aggregate-write sites | ✅ | saveTeam/deleteTeam intentionally not wired (architecturally correct — per-team files, not aggregate; aggregate guard happens at TeamsBackupService.updateBackup) |
| A2 N=30 ring buffer + atomic index | ✅ | 25/25 service + 19 controller/routes |
| A2 GET history + POST restore/:slot | ✅ | API path shape is `/api/teams/backup/...` not spec's `/api/teams-backup/...` (cosmetic) |
| C1 invariant + override | ✅ | 9 type tests + 7 integration tests + 1 e2e |
| e2e artificial nuke ship gate | ✅ | f91cd608 — 5-phase + override + in-flight wipe regression |
| No regressions in existing tests | ✅ | 79/79 P0 suite green |

---

### Recommendation

**Merge as-is.** Three independent layers, each tested independently, e2e ties them together. The 2026-05-03 failure mode is now caught at three points (write boundary via A1, snapshot durability via A2, boot detection via C1) and the e2e regression test pins all three.

1. **One-line spec/PR clarification** that `saveTeam`/`deleteTeam` are correctly NOT wired due to per-team-file architecture (avoid future "is this a gap?" question).
2. **Optional P1**: extend A2 rotating-snapshot pattern to `projects.json` and `runtime/*.json` (spec §9 Q2 already flagged this).
3. **Optional P1**: extend C1 predicate to also consult rotating-slot max count (spec §9 Q4 — defense-in-depth even if A1 fails to fire).
4. **Optional**: monitor `~/.crewly/teams-backup-history/` directory size for heavy installs (30 slots × full snapshot can grow with many teams + large prompts; bounded but worth a metric).

PR #415, opened 2026-05-03
feat(memory): M1 — auto-inject mission/OKR context into agent prompts

stevehuang0115:
## Summary
- New `MissionContextService` (singleton) assembles a 1-2KB markdown card from goals + active missions + optional sub-OKR + recent review summaries
- `PromptGeneratorService.generateMemberPrompt()` now injects that card between the Skills block and the Context Reminder footer
- Hard 2KB cap inside the service; builder hook is fail-soft (any read error logs and skips, so prompt assembly never breaks)
- Empty cards emit nothing — zero-signal teams unchanged

## Why
M1 (Memory Phase 1) of the agent memory roadmap: every agent prompt should carry current mission/OKR context so agents reason against goals without manual reminders.

## Files
- `backend/src/services/memory/mission-context.service.ts` (new) — assembly + 2KB cap
- `backend/src/services/memory/mission-context.service.test.ts` (new) — 27 unit + integration tests
- `backend/src/services/prompt/prompt-generator.service.ts` — adds import + hook between Skills and Context Reminder
- `backend/src/services/prompt/prompt-generator.service.test.ts` — adds `MissionContextService` mock + 6 M1 cases

## Coordination
Confirmed zero overlap with Sam's P0-1 work (`prompt-builder.service.ts` + `prompt-modules/soul.module.ts` — different code path entirely).

## Test plan
- [x] Backend typecheck (`npx tsc --noEmit -p backend`) clean
- [x] `mission-context.service.test.ts` — 27 pass
- [x] `prompt-generator.service.test.ts` — 6 new M1 cases pass; legacy cases unchanged
- [x] Total: 33 / 33 pass
- [ ] CI green (pending)

## Behavior
- **Position**: Skills → **Mission Context (M1)** → Context Reminder footer
- **Empty signals** (no goals, no missions): card is empty string → no section emitted, prompt unchanged
- **Read failure**: logger.warn, no throw, prompt completes without the card
- **Cap**: hard truncate at 2KB by recency-then-importance inside the service

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #421, opened 2026-05-04
fix(memory): #415 — rewire M1 mission-context to PromptBuilderService (live path)

stevehuang0115:
## Summary

Closes the dead-code-path bug Arch flagged on PR #415. The original M1 implementation hooked `PromptGeneratorService.generateMemberPrompt`, which is no longer in the agent boot path — the live path is `PromptBuilderService` → `PromptAssemblyService.modules[]` (the prompt-modules pattern Sam landed in P0-1 Phase 2). Result: the mission card was never reaching production prompts.

This PR brings M1 back via the correct mechanism and addresses Arch's three secondary findings.

## What changed

- **`MissionContextModule`** (NEW, `backend/src/services/ai/prompt-modules/`) implementing the standard `PromptModule` interface. Sits at priority 7 — after team references, before learning references — so the agent first sees who they work with, then which mission they serve. `compactable=true`, `maxTokens=600`. Build is fail-soft: any service error logs a warn and returns `''`.
- **Registered** the module in `PromptAssemblyService.registerDefaultModules()`.
- **`MissionContextService`** migrated from #415 with the `ownerTeamId === teamId` filter widened — accepts `ancestorTeamIds[]` and matches missions whose owner is anywhere in the team-graph scope (own team ∪ ancestors). Caller resolves the chain; the service does no graph traversal of its own.
- **`ModuleConfig`** gains `teamSlug` and `teamAncestorIds`. `PromptBuilderService.buildModuleConfigFromTeamMember` now populates both — `teamSlug` via local `toTeamSlug` (mirrors `template.controller.toSlug` convention), `teamAncestorIds` via `team.parentTeamId` (single-level today; deeper ancestry deferred to the runtime team-graph resolver).
- **No `process.cwd()`** — module consumes `config.projectPath` the assembler already plumbs.
- **Did NOT** bring over the `prompt-generator.service.ts` hook from #415 — that file stays untouched on this branch (it's the dead path).

## Arch finding mapping

| Finding | Resolution |
|---|---|
| Wired to dead path (PromptGeneratorService) | Replaced with PromptModule registered in PromptAssemblyService (live path) |
| Missing teamSlug pass-through | `teamSlug` now in `ModuleConfig`; populated from `team.name` via `toTeamSlug` |
| Hardcoded `process.cwd()` | Module reads `config.projectPath` exclusively; tested |
| `ownerTeamId`-only filter | Service accepts `ancestorTeamIds[]`, scope-set lookup includes parents |

## Test plan

- [x] `mission-context.module.test.ts` (NEW, 13) — metadata, shouldInclude (teamId / evalMode), payload forwarding, no process.cwd usage, trim/empty/throw handling
- [x] `mission-context.service.test.ts` (21) — original 19 + 2 new (ancestor inclusion, de-dup overlap)
- [x] `prompt-builder.service.test.ts` (+4) — slug + ancestor population in `buildModuleConfigFromTeamMember`
- [x] `prompt-assembly.service.test.ts` (updated) — module count 17 → 18, `mission_context` registered
- [x] **All 173 affected tests pass** in 4 suites
- [x] `tsc --noEmit -p backend` clean
- [ ] Manual smoke: spawn an agent with a populated `.crewly/missions/` directory and verify the mission card lands in the assembled system prompt

## Coordination

Ping sent to Max (`crewly-product-max-358c7cb7`) to follow the same pattern for #416 (M2 working-memory wake card). Plumbing added here (`teamSlug`, `teamAncestorIds`) lands first; M2's module should rebase on this branch and consume the same `ModuleConfig` fields.

## Spec / issue refs

- #415 — M1 dead-path issue (this PR replaces)
- `.crewly/specs/2026-05-03-state-persistence-fix-spec.md` — unrelated; just to confirm we're following the same pattern as Sam's P0-1 Phase 2

### DIFF
--- a/backend/src/services/ai/prompt-builder.service.test.ts
+++ b/backend/src/services/ai/prompt-builder.service.test.ts
@@ -726,6 +726,35 @@ describe('PromptBuilderService', () => {
 			});
 			expect(config.subordinates?.[0].name).toBe('Pre-resolved');
 		});
+
+		// Mission/OKR card plumbing — fix #415 secondary findings.
+		// Slug is required to resolve `<slug>-team-sub-okr.md`; ancestor
+		// chain is required for the OKR-hierarchy filter widening.
+		describe('mission-context plumbing (#415)', () => {
+			it('slugifies team.name into teamSlug (lowercase, hyphenated, alnum-only)', () => {
+				const team = { ...buildTeam(), name: 'Crewly Product!' } as Team;
+				const config = buildModuleConfigFromTeamMember(buildMember(), team, runtime);
+				expect(config.teamSlug).toBe('crewly-product');
+			});
+
+			it('teamSlug is undefined when team.name is empty', () => {
+				const team = { ...buildTeam(), name: '' } as Team;
+				const config = buildModuleConfigFromTeamMember(buildMember(), team, runtime);
+				expect(config.teamSlug).toBeUndefined();
+			});
+
+			it('teamAncestorIds contains parentTeamId when set', () => {
+				const team = { ...buildTeam(), parentTeamId: 'parent-team-x' } as Team;
+				const config = buildModuleConfigFromTeamMember(buildMember(), team, runtime);
+				expect(config.teamAncestorIds).toEqual(['parent-team-x']);
+			});
+
+			it('teamAncestorIds is undefined when team has no parent', () => {
+				const team = { ...buildTeam(), parentTeamId: undefined } as Team;
+				const config = buildModuleConfigFromTeamMember(buildMember(), team, runtime);
+				expect(config.teamAncestorIds).toBeUndefined();
+			});
+		});
 	});
 
 	describe('buildSystemPromptWithTeamContext (WIRE-2)', () => {
--- a/backend/src/services/ai/prompt-builder.service.ts
+++ b/backend/src/services/ai/prompt-builder.service.ts
@@ -157,7 +157,30 @@ export function buildModuleConfigFromTeamMember(
 		teamQualityGate: team.qualityGate,
 		serviceContract: team.serviceContract,
 		teamOwnershipScope: team.ownershipScope,
+
+		// === Team-graph context (consumed by MissionContextModule) ===
+		// Slug derived from team name; matches the convention used elsewhere
+		// (e.g. template.controller.toSlug + sessionName composition).
+		// Used to resolve `<teamSlug>-team-sub-okr.md` under .crewly/goals/.
+		teamSlug: team.name ? toTeamSlug(team.name) : undefined,
+		// Single-level parent only — modules that need deeper ancestry
+		// should call a graph resolver. Walking the full parent chain
+		// here would require runtime team-graph access this builder
+		// intentionally does not have.
+		teamAncestorIds: team.parentTeamId ? [team.parentTeamId] : undefined,
 	};
+}
+
+/**
+ * Convert a team name to a URL-safe slug (lowercase, hyphens, alnum only).
+ * Mirrors `toSlug` in `controllers/template/template.controller.ts` so
+ * goal / OKR file naming stays consistent across the codebase.
+ *
+ * @param name - Human-readable team name
+ * @returns Lowercase slug suitable for file path segments
+ */
+function toTeamSlug(name: string): string {
+	return name.toLowerCase().replace(/\s+/g, '-').replace(/[^a-z0-9-]/g, '');
 }
 
 /**
--- a/backend/src/services/ai/prompt-modules/index.ts
+++ b/backend/src/services/ai/prompt-modules/index.ts
@@ -26,6 +26,7 @@ export { TeamReferenceModule } from './team-reference.module.js';
 export { ProjectReferenceModule } from './project-reference.module.js';
 export { UserProfileReferenceModule } from './user-profile-reference.module.js';
 export { LearningReferenceModule } from './learning-reference.module.js';
+export { MissionContextModule } from './mission-context.module.js';
 export { CommunicationModule } from './communication.module.js';
 export { RecoveryModule } from './recovery.module.js';
 export { LifecycleModule } from './lifecycle.module.js';
--- a/backend/src/services/ai/prompt-modules/mission-context.module.test.ts
+++ b/backend/src/services/ai/prompt-modules/mission-context.module.test.ts
@@ -1,0 +1,170 @@
+/**
+ * Tests for MissionContextModule — the live wire-in for mission/OKR
+ * auto-injection (fix #415, replaces the dead PromptGeneratorService
+ * hook).
+ *
+ * @module services/ai/prompt-modules/mission-context.module.test
+ */
+
+import { describe, it, expect, beforeEach, jest } from '@jest/globals';
+import { MissionContextModule } from './mission-context.module.js';
+import type { ModuleConfig } from './prompt-module.interface.js';
+import { MissionContextService } from '../../memory/mission-context.service.js';
+
+/**
+ * Build a minimal MissionContextService stub that records every call
+ * and returns whatever the test sets up.
+ */
+function makeStubService(returnValue: string | Error): {
+	service: MissionContextService;
+	calls: Array<Parameters<MissionContextService['getContextForAgent']>[0]>;
+} {
+	const calls: Array<Parameters<MissionContextService['getContextForAgent']>[0]> = [];
+	const service = {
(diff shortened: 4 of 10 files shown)


stevehuang0115 (commented):

Re-review of #421 against the four findings I left on #415. Net: ship it. All four fixes land cleanly, the new wire-in is on the live path, and the test plan pins the regression frontier where it matters.

---

### Finding-by-finding closure

| #415 finding | #421 resolution | Verdict |
|---|---|---|
| **Wired to dead `PromptGeneratorService` path** — mission card never reached production prompts | New `MissionContextModule` registered in `PromptAssemblyService.registerDefaultModules()` at line 118; no touch to `prompt-generator.service.ts`. Dead path correctly left dead. | ✅ Closed |
| **Missing `teamSlug` pass-through** — service couldn't resolve `<slug>-team-sub-okr.md` | `teamSlug` added to `ModuleConfig` (`prompt-module.interface.ts:97-103`); `PromptBuilderService.buildModuleConfigFromTeamMember` populates via local `toTeamSlug` (`prompt-builder.service.ts:163-181`). Empty-name → `undefined` test pins the guard. | ✅ Closed |
| **Hardcoded `process.cwd()`** | Module reads `config.projectPath` exclusively; defensive `if (!config.projectPath) return ''` at `mission-context.module.ts:99`; explicit `expect(calls[0].projectPath).not.toBe(process.cwd())` regression test. | ✅ Closed |
| **`ownerTeamId === teamId` only filter** — sub-team agents miss parent OKRs | Service accepts `ancestorTeamIds[]`, builds `Set([teamId, ...ancestors])` via `uniqueNonEmpty` (`mission-context.service.ts:206`), filters with `scope.has(parsed.ownerTeamId)` (`mission-context.service.ts:259`); builder populates `team.parentTeamId ? [parentTeamId] : undefined`. Both ancestor-inclusion and de-dup tests present. | ✅ Closed |

---

### Module-layer review (the new live wire-in)

`MissionContextModule` (`backend/src/services/ai/prompt-modules/mission-context.module.ts`):

- **Interface compliance** — implements `PromptModule`: `name`, `priority`, `maxTokens`, `compactable`, `shouldInclude`, `build`. Matches sibling reference-family modules (`team_references`, `learning_references`, `memory_references`).
- **Priority slot=7** is correct: sits between `team_references` (priority=6) and `learning_references` (priority=10). Comment justifies the ordering ("agent first sees who they work with, then which mission they serve, then how to learn from past work") — defensible narrative.
- **Naming convention** (`mission_context`, snake_case) matches the *_references family in this directory (vs. kebab-case for non-reference modules like `role-boundary`, `team-norms`). Convention is split here but the choice is internally consistent. No change needed.
- **`shouldInclude` guard** correctly skips on missing `teamId` (no scoping anchor) and `evalMode=true` (token saving). Defensive for direct callers, with a redundant guard in `build()` itself — belt-and-suspenders, fine.
- **Fail-soft `build()`** (lines 91-110): single try/catch wrapping the service call; any throw logs a warn and returns `''`. The assembler can't be wedged by a malformed missions directory. This is the correct posture for an auto-injected card.
- **Constructor seam** (line 65, `service?: MissionContextService`) — clean test injection while preserving singleton ergonomics in production. The 13 module-level tests use this seam well.

---

### Builder-layer review (`buildModuleConfigFromTeamMember`)

`prompt-builder.service.ts:160-170`:

- **`teamSlug`**: `team.name ? toTeamSlug(team.name) : undefined`. Truthy check correctly maps empty string → `undefined` (test at line 740). The local `toTeamSlug` (lines 178-181) is a 4-line lowercase + `\s→-` + alnum-only filter.
- **`teamAncestorIds`**: `team.parentTeamId ? [team.parentTeamId] : undefined`. Single-level only. Doc-comment is honest about the limitation: *"single-level today; deeper ancestry deferred to the runtime team-graph resolver."*
- **Tests** (lines 729-756) cover the four shape combinations: slugify, empty-name guard, parent-present, parent-absent. Good shape.

**Spec deviation worth naming (non-blocker):** The original M1 finding was that the filter should match the team-graph scope. `team.parentTeamId` only walks one level up. For deeper Crewly team hierarchies (Mia→Sam→Leo style: 3 levels), a Leo agent will see his immediate team's missions and Sam's team's missions, but NOT Mia's. The PR correctly documents this as deferred to a "runtime team-graph resolver." **Accept the trade-off** — single-level covers the dominant case (parent OKR inheritance), and the service-side `ancestorTeamIds` interface is generic enough to absorb deeper resolution without an API change. Worth a P1 follow-up to land the multi-level walker.

---

### Service-layer review

`MissionContextService` (`backend/src/services/memory/mission-context.service.ts`):

- **Pure-helper extraction** is excellent: `extractMeaningfulLines`, `collectRecentStatusUpdates`, `renderMissionCard`, `enforceHardCap`, `uniqueNonEmpty` all exported and individually tested. The hard-cap test that proves it stays ≤2KB even with a 1000-line stuffed `goals.md` is a good regression pin against future renderer drift.
- **Ancestor scope-set** (line 205-206): `uniqueNonEmpty([teamId, ...ancestorTeamIds])` correctly handles dedup, falsy entries, and order preservation. Caller test passes `['team-A', 'team-A']` to verify the dedup.
- **Active-only filter**: `parsed.status === 'active'` — tested explicitly with active/paused/other-team triplet. Correct.
- **Corrupt JSON tolerance** (line 277-282): `try/catch` per file, debug-level log, skips silently. Test at line 472-491 (`'treats corrupt mission JSON as a missing mission, not a failure'`) is a direct regression pin for the dominant operational hazard. ✅
- **Missing-directory tolerance** (line 250-253): `fs.readdir` failure → `[]`, comment correctly notes "missing == no missions == empty section, not an error."
- **Hard cap enforcement** truncates from the bottom with a labeled ellipsis (`'… (truncated to fit 2KB cap)'`) — preserves heading + project OKR + active missions when under pressure (correct prioritization since recent status updates are the most replaceable).

---

### Assembly-layer review

`prompt-assembly.service.ts:118` — `new MissionContextModule()` registered in `registerDefaultModules` between `TeamReferenceModule` and `CapabilityOverlayModule`. Module count assertion bumped 17→18 in the assembly test.

The registration order in `registerDefaultModules` is informational — actual emit order is governed by `priority` — but it does set the construction order, and the inline comment correctly anchors the rationale.

---

### Observations (non-blocking, none required for merge)

1. **`toTeamSlug` duplication.** This PR's local `toTeamSlug` mirrors `template.controller.toSlug`. Two implementations, same intent — drift risk. Suggest extracting to `backend/src/utils/team-slug.utils.ts` in a follow-up so future slug-rule changes (e.g. unicode handling) don't have to be made twice.
2. **Multi-level ancestor walking** (above) — track as P1; current PR's interface absorbs the eventual fix without breaking changes.
3. **Default `priority='medium'` fallback** in `formatMissionLine` (`mission-context.service.ts:399`) — fine, but worth surfacing in the Mission type as a default rather than a renderer-side fallback to keep the contract close to the data.
4. **Manual smoke test still unchecked** in PR description (last test-plan checkbox: spawn an agent + populated `.crewly/missions/` and verify card lands in assembled prompt). Tests cover the unit boundary; would be reassuring to confirm one end-to-end render before merge, but I'd accept this as a post-merge smoke since 173/173 deterministic tests already exercise every layer of the path.

---

### Test plan adequacy

- 13 module tests — interface, guards, payload forwarding, fail-soft, no-cwd
- 21 service tests — original 19 + 2 new (ancestor inclusion, dedup)
- 4 builder tests — slug + ancestor population at the construction site
- Assembly registration test — count bump + name presence

Coverage hits every layer of the new wire-in plus every behavior change in the migrated service. Real-fs roundtrip in service tests catches the failure modes that mocks would miss (corrupt JSON, missing directories, hard-cap truncation against actual byte length). 173/173 + tsc clean.

---

### Recommendation

**Merge as-is.** All four #415 findings closed cleanly, the wire-in is on the live `PromptAssemblyService` pipeline (verified — no orphan code path), and the test plan covers the regression frontier. The single pending checkbox (manual smoke) can land as a post-merge sanity check given the comprehensiveness of the deterministic suite.

1. **P1**: replace single-level `team.parentTeamId` with a multi-level team-graph walker so 3+ deep hierarchies inherit grandparent OKRs.
2. **P2**: extract `toTeamSlug` to a shared util to prevent drift with `template.controller.toSlug`.
3. **P2**: verify Max's M2 working-memory module (PR #416 fix) consumes `teamSlug` + `teamAncestorIds` via the same `ModuleConfig` plumbing landed here.

PR #423, opened 2026-05-04
feat(memory): M2 — working-memory wake card module (real prompt path)

stevehuang0115:
## Summary

Adds the "Why You Are Awake" **wake card** to the **live** prompt assembly path (`PromptAssemblyService`) at priority 8 — Memory Phase 1 fix M2.

- The card answers, in ≤1.5KB: *what triggered me*, *what I should do next*, *what's blocking*. Built from the agent's active `WorkItem`, its parent `Request` chain, and any open loops (blockers / pending checkpoints).
- Lives **after** the slot reserved for Leo's M1 Mission Card (priority 7), so the agent reads short-term "why-now" runtime context after the long-term mission frame.
- Wired into `PromptAssemblyService.registerDefaultModules()` — the **same** real path Sam used for P0-1 Phase 2 (team-leader soul + role-boundaries) and the path Leo's M1 PR re-targets. The previous draft injection point (`PromptGeneratorService.generateMemberPrompt`) is dead code at runtime; this PR avoids that path entirely.

## Architecture note — dead-code-path defect

`PromptGeneratorService.generateMemberPrompt` is **not** the live runtime prompt builder. Every production agent is built via `PromptBuilderService` → `PromptAssemblyService`. Wake card now sits inside that real chain. Leo's M1 PR makes the same correction for the mission card.

## Components

| File | Role |
|---|---|
| `backend/src/services/memory/working-memory.service.ts` | Singleton: load active WorkItems, build Request chain, collect open loops, render ≤1.5KB markdown wake card with hard byte-cap at line boundaries. |
| `backend/src/services/ai/prompt-modules/working-memory.module.ts` | Thin `PromptModule` adapter (priority 8, maxTokens 400, compactable **false**). Fail-soft: any service error is logged and converted to `''`. |
| `backend/src/services/ai/prompt-modules/prompt-assembly.service.ts` | Registers `WorkingMemoryModule` in default module set after `CapabilityOverlayModule`. |
| `backend/src/services/ai/prompt-modules/index.ts` | Re-export. |

### Why `compactable = false`

The wake card is already hard-capped at `WAKE_CARD_HARD_CAP_BYTES = 1536` inside the service. The assembler's 50% trim heuristic would cut the structured "Required next action" / "Known constraints" sections mid-line — destroying the agent's actionable context. Better to emit the full card or skip it via the empty return.

### Why budget = 400 tokens

Service caps card bytes at 1536; `1536 / 4 ≈ 384` tokens. Round to 400 for headroom (markdown heading + assembler separators).

## Test plan

- [x] `working-memory.module.test.ts` — 13 cases: metadata, `shouldInclude` gating (sessionName + projectPath/projectRoot fallback), `build()` delegation contract (memberId fallback, projectRoot fallback, teamId pass-through), fail-soft on service throw, `PromptAssemblyService` registration + ordering vs `team_references`.
- [x] `working-memory.service.test.ts` — 46 cases: card data collection, hard-cap enforcement at line boundaries (regex updated to permit real `## hdr` headings), public `getWakeCardForAgent` contract.
- [x] `npx jest backend/src/services/ai/prompt-modules/working-memory.module.test.ts backend/src/services/memory/working-memory.service.test.ts --no-coverage` → **59/59 pass**
- [x] `npx tsc --noEmit -p backend/tsconfig.json` clean
- [ ] Reviewer: confirm no overlap with Leo's M1 `ModuleConfig` shape (none expected — M2 only consumes existing fields: `sessionName`, `memberId`, `projectPath`, `projectRoot`, `teamId`).

## Coordination

- Independent branch from Leo's M1 PR. **No merge race** — neither PR touches the other's module file or the `ModuleConfig` schema.
- Ordering comment in `prompt-assembly.service.ts` documents the priority-7 slot Leo's PR fills.

### DIFF
--- a/backend/src/services/ai/prompt-modules/index.ts
+++ b/backend/src/services/ai/prompt-modules/index.ts
@@ -30,6 +30,7 @@ export { CommunicationModule } from './communication.module.js';
 export { RecoveryModule } from './recovery.module.js';
 export { LifecycleModule } from './lifecycle.module.js';
 export { RoleBoundaryModule } from './role-boundary.module.js';
+export { WorkingMemoryModule } from './working-memory.module.js';
 
 // Context loaders
 export {
--- a/backend/src/services/ai/prompt-modules/prompt-assembly.service.ts
+++ b/backend/src/services/ai/prompt-modules/prompt-assembly.service.ts
@@ -24,6 +24,7 @@ import { CapabilityOverlayModule } from './capability-overlay.module.js';
 import { DomainSOPModule } from './domain-sop.module.js';
 import { RiskPolicyModule } from './risk-policy.module.js';
 import { TeamNormsModule } from './team-norms.module.js';
+import { WorkingMemoryModule } from './working-memory.module.js';
 
 /**
  * Default total token budget for all prompt modules combined.
@@ -115,6 +116,11 @@ export class PromptAssemblyService {
 			new SkillsReferenceModule(),
 			new TeamReferenceModule(),
 			new CapabilityOverlayModule(),
+			// Mission Card (priority 7) is registered by Leo's M1 dead-code-path
+			// rewire — it lands here once that PR merges. Wake card (priority 8)
+			// must follow the mission card so the agent reads short-term
+			// "why-now" runtime context after the long-term mission frame.
+			new WorkingMemoryModule(),
 			new ProjectReferenceModule(),
 			new CommunicationModule(),
 			new DomainSOPModule(),
--- a/backend/src/services/ai/prompt-modules/working-memory.module.test.ts
+++ b/backend/src/services/ai/prompt-modules/working-memory.module.test.ts
@@ -1,0 +1,279 @@
+/**
+ * Unit tests for WorkingMemoryModule.
+ *
+ * Covers:
+ *   - Module metadata (name / priority / maxTokens / compactable).
+ *   - shouldInclude gating on sessionName + projectPath/projectRoot.
+ *   - build() delegates to WorkingMemoryService.getWakeCardForAgent
+ *     with the correct request shape.
+ *   - Fail-soft: service errors are swallowed and empty string is
+ *     returned so prompt assembly never breaks.
+ *   - Empty service result collapses to '' (assembler then drops the
+ *     module entry instead of emitting an empty heading).
+ *   - Registration in PromptAssemblyService default modules — guards
+ *     against accidental removal during future module reshuffles.
+ *
+ * @module services/ai/prompt-modules/working-memory.module.test
+ */
+
+import { WorkingMemoryModule } from './working-memory.module.js';
+import { ModuleConfig } from './prompt-module.interface.js';
+import { WorkingMemoryService } from '../../memory/working-memory.service.js';
+import { PromptAssemblyService } from './prompt-assembly.service.js';
+
+describe('WorkingMemoryModule', () => {
+  let module: WorkingMemoryModule;
+
+  const baseConfig: ModuleConfig = {
+    sessionName: 'crewly-product-max-358c7cb7',
+    memberId: '358c7cb7-eb20-482a-87b2-655eb8cdde4e',
+    role: 'developer',
+    teamId: '28a4ac10-f37f-4286-ad8c-6926a0e177f5',
+    projectPath: '/Users/me/projects/crewly',
+    agentSkillsPath: '/path/to/skills/agent',
+    tlSkillsPath: '/path/to/skills/team-leader',
+    projectRoot: '/Users/me/projects/crewly',
+  };
+
+  beforeEach(() => {
+    WorkingMemoryService.clearInstance();
+    module = new WorkingMemoryModule();
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+    WorkingMemoryService.clearInstance();
+  });
+
+  // -------------------------------------------------------------------------
+  // Module metadata
+  // -------------------------------------------------------------------------
+
+  describe('metadata', () => {
+    it('declares name="working-memory"', () => {
+      expect(module.name).toBe('working-memory');
+    });
+
+    it('declares priority=8 — after MissionCard (7) and TeamReference (6)', () => {
+      expect(module.priority).toBe(8);
+    });
+
+    it('declares maxTokens=400 — covers the 1.5KB hard cap with margin', () => {
+      // Service caps card at 1536 bytes; 1536/4 ≈ 384 tokens. Module budgets
+      // 400 to leave headroom for the markdown heading and assembler
+      // separators added during composition.
+      expect(module.maxTokens).toBe(400);
+      expect(module.maxTokens).toBeGreaterThanOrEqual(384);
+    });
+
+    it('declares compactable=false — wake card must not be 50%-trimmed', () => {
+      // The wake card carries structured "Required next action" / "Known
+      // constraints" sections. The assembler's 50% trim heuristic would cut
+      // them mid-line and destroy the agent's actionable context. Better to
+      // emit the full card or skip it (via empty return) than to truncate.
+      expect(module.compactable).toBe(false);
+    });
+  });
+
+  // -------------------------------------------------------------------------
+  // shouldInclude gating
+  // -------------------------------------------------------------------------
+
+  describe('shouldInclude', () => {
+    it('returns true for a fully-populated config', () => {
+      expect(module.shouldInclude(baseConfig)).toBe(true);
+    });
+
+    it('returns false when sessionName is empty', () => {
+      const cfg: ModuleConfig = { ...baseConfig, sessionName: '' };
+      expect(module.shouldInclude(cfg)).toBe(false);
+    });
+
+    it('returns false when both projectPath and projectRoot are missing', () => {
+      const cfg: ModuleConfig = {
+        ...baseConfig,
+        projectPath: undefined,
+        projectRoot: '',
+      };
+      expect(module.shouldInclude(cfg)).toBe(false);
+    });
+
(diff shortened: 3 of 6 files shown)


stevehuang0115 (commented):

Standard review of #423 against three lenses: (a) Steve's explicit dead-code-path verification ask, (b) the live wire-in shape vs. Leo's #421 mirror, (c) module + service quality. Net: ship it.

---

### Live-path verification — the Steve-flagged check

This was the explicit verify-this from the dispatch ("同 dead-code-path bug 你在 #415 找过, Max 这条已经 mirror Leo 的 #421 fix pattern 接 PromptBuilderService, 你 verify 这点").

**Confirmed: PASS.**

| Signal | Evidence |
|---|---|
| Module wired into LIVE assembler | `prompt-assembly.service.ts:120` registers `new WorkingMemoryModule()` inside `registerDefaultModules()` after `CapabilityOverlayModule` |
| LIVE assembler is the production path | `PromptBuilderService` (at `services/ai/prompt-builder.service.ts`) consumes `PromptAssemblyService.getModules()` via `buildSystemPromptWithTeamContext` (line 526) — this is the Sam P0-1 Phase 2 path the team-leader soul + role-boundaries also flow through |
| DEAD path untouched | `PromptGeneratorService.generateMemberPrompt` at `services/prompt/prompt-generator.service.ts` not modified — it remains the orphan since P0-1 Phase 2, just like in #421 |
| Mirrors Leo's #421 pattern exactly | Same registration site (`registerDefaultModules`), same constructor seam, same fail-soft try/catch, same `priority`-driven ordering with explicit reservation comment for the sibling slot |
| Reservation comment for Leo's slot | Inline comment at the registration site documents that priority-7 mission card ships from #421 and that wake card (priority 8) intentionally follows it so the agent reads long-horizon mission context before short-horizon "why-now" wake context — coordination preserved |

The dead-code-path bug from #415 is actively avoided. Same fix posture as Leo's #421.

---

### Module-layer review

`WorkingMemoryModule` (`backend/src/services/ai/prompt-modules/working-memory.module.ts`):

- **Interface compliance** — `name`, `priority`, `maxTokens`, `compactable`, `shouldInclude`, `build` all present. Matches sibling shape.
- **`compactable=false` is the correct call.** Service hard-caps at 1536 bytes; if the assembler's 50% trim heuristic kicked in, it would shred the structured "Required next action" / "Known constraints" sections mid-line. Better to emit the full structured card or skip via empty return — partial card is worse than no card. Documented in the module comments.
- **`priority=8` is correct.** Lands after Leo's M1 (priority 7) so agents read mission context before runtime context. Lands after team_references (6) and capability overlays so the agent knows the team and role frame before the "what's-on-your-plate-right-now" overlay.
- **`shouldInclude` guards** — skip on missing `sessionName`; skip on missing both `projectPath` and `projectRoot`. Belt-and-suspenders with the inner `build()` check that returns `''` if both paths missing. Defensive without being noisy.
- **`agentId` fallback** — uses `config.memberId` for log-correlation when present, falls back to `sessionName` for legacy SessionConfig callers. Good forward-back compat without leaking the choice into the data shape.
- **`projectRoot` ↔ `projectPath` fallback** — `config.projectPath ?? config.projectRoot` covers the dual-key reality of ModuleConfig today. Not pretty, but correct given the existing interface; worth a P3 cleanup at the interface level eventually.
- **Fail-soft `build()`** — single try/catch wraps the service call; any throw logs warn + returns `''`. Assembler can't be wedged by a malformed task pool or request store. Same posture as Leo's #421.

---

### Service-layer review

`WorkingMemoryService` (`backend/src/services/memory/working-memory.service.ts`):

- **Pure-helper extraction is excellent.** `pickCurrentWorkItem`, `buildUpstreamChain`, `collectOpenLoops`, `summarizeWorkItem`, `renderWakeCard`, `enforceHardCap` all exported and individually unit-tested. Same pattern as `MissionContextService` in #421 — consistency across the M1/M2 family is good for future maintainers.
- **`pickCurrentWorkItem`** — filters by `target===sessionName` + status in `ACTIVE_WORK_ITEM_STATUSES` (proposed/accepted/running/blocked/escalated; terminal statuses excluded). Picks max(`startedAt ?? createdAt`) for tiebreak. Reasonable selection: most recently activated active item. Worth flagging that two items started in the same millisecond will tie-break by array order — extremely unlikely in practice.
- **`ACTIVE_WORK_ITEM_STATUSES`** — set membership over a fixed list, cleanly typed via `ReadonlySet<WorkItemStatus>`. The terminal-status exclusion test pins the regression frontier (verified/cancelled/failed/rejected/done all skipped).
- **`buildUpstreamChain` cycle guard** — `visited: Set<string>` walks `parentWorkItemId`; bails on revisit, bails on `null/undefined`, bails on missing parent in pool, bails on `maxItems`. Returns reversed for chronological order. The cycle test (p↔q anchored at current.parentWorkItemId='p' expects ['q','p']) is a real regression pin against future graph mutations introducing accidental cycles.
- **`collectOpenLoops` is intentionally thin.** Today only emits "Awaiting user confirmation: <reason>" when `requiresConfirmation && confirmationReason`. Comment correctly flags this as an M5/M6 extension point (decisions, lessons, outstanding sub-tasks). Forward-compatible without overcommitting now — accept.
- **`enforceHardCap` truncation** — line-boundary truncation preserves rendered structure; falls back to byte-truncation only if the header alone exceeds the cap (a degenerate case). 5000-char description test asserts ≤1536 bytes — pins the contract. Same shape as MissionContext's hard-cap, which leads to one observation (below).
- **`teamId` in `WakeCardRequest`** — accepted but currently unused. Comment says "reserved for future filters." Forward-compat hook vs. dead parameter is a judgment call; I'd lean toward accept given the cost of changing the request shape later, but it's a fair P3 to either wire it up or strip it.
- **`TaskPoolService.getInstance().getAllItems()` + `RequestService.getInstance(projectPath).getById(requestId)`** — both leverage existing singleton + clearInstance() seam, consistent with Leo's M1 pattern. Test injection works; production callers don't need to know about the seam.

---

### Test plan adequacy

- **13 module tests** — metadata, shouldInclude (4 cases including projectRoot fallback), build delegation (memberId fallback, projectRoot fallback, teamId pass-through), fail-soft on service throw, defensive-empty-on-no-paths, registration in PromptAssemblyService, ordering vs `team_references` (snake_case sibling)
- **46 service tests** — pure-helper coverage (pickCurrentWorkItem, cycle guard, terminal-status exclusion, collectOpenLoops thin path), real-fs RequestService roundtrip with temp dir, spied TaskPool, fail-soft on TaskPool throw, missing-parent-Request graceful downgrade, hard-cap byte-length assertion against 5000-char stuff

The cycle test is the kind of thing reviewers love to see — a deliberate adversarial case (p↔q anchored at current.parentWorkItemId='p') that would silently infinite-loop if the visited Set were dropped. The hard-cap stuff test is a similar regression pin against future renderer changes that might accidentally bloat the card.

59/59 tests pass; typecheck clean; no console.log noise added.

---

1. **`enforceHardCap` duplication.** Both `MissionContextService` (#421, 2048-byte cap) and `WorkingMemoryService` (this PR, 1536-byte cap) implement nearly identical line-boundary truncation logic. Two implementations, same intent — drift risk if one gets a fix the other doesn't. Suggest extracting to `backend/src/services/memory/utils/hard-cap.util.ts` (or similar) in a P2 follow-up. Cap value stays per-service; helper just takes `(text, capBytes)`.

2. **Naming convention split.** `working-memory` (kebab) vs `mission_context` (snake) for sibling modules. Both are internally consistent within their families (kebab for Sam's P0-1 Phase 2 modules like `role-boundary` / `team-norms`, snake for the `*_references` and `*_context` family). Not a regression — just worth flagging that the prompt-modules directory is now mixed-convention. P3.

3. **`teamId` in `WakeCardRequest` currently unused.** Reserved for future filters per code comment. Either wire it up in M3+ (e.g., scope wake card to team-bound tasks only) or strip it from the request shape. P3 — accept the forward-compat hook for now.

4. **`pickCurrentWorkItem` tiebreaker on equal timestamps.** Two items started in the same millisecond will tie-break by array order from `getAllItems()`. Vanishingly unlikely in production but technically nondeterministic. Could add a `id` lex tiebreaker if it ever surfaces. P3.

5. **`projectRoot ?? projectPath` interface duality.** Both keys exist in `ModuleConfig` and both module + service support them via `??` fallbacks. Worth a P2 cleanup at the `ModuleConfig` interface level: pick one canonical key, deprecate the other, migrate callers. Not Max's problem on this PR.

6. **`collectOpenLoops` will need a real spec for M5/M6.** Today it's a one-line "awaiting confirmation" emitter. By the time M5 (decisions) and M6 (lessons) land, the renderer needs to know what fits in the 1536-byte cap when multiple loop signals compete for space. Worth a P1 spec note before adding more emitters so the priority order is explicit (e.g., confirmation > blocking decisions > outstanding sub-tasks > lessons).

---

### Recommendation

**Merge as-is.** Live-path wire-in mirrors Leo's #421 pattern exactly and avoids the dead-code-path bug from #415. Module is fail-soft, priority slot is coordinated with M1, hard-cap + cycle-guard + terminal-status-exclusion all have explicit regression tests. 59/59 deterministic suite pins the contract.

1. **P2**: extract `enforceHardCap` to a shared util to prevent drift with MissionContext's version.
2. **P2**: pick one canonical key for project path on `ModuleConfig` (`projectPath` vs `projectRoot`), deprecate the other, migrate callers.
3. **P1**: spec the `collectOpenLoops` priority order before M5/M6 add more loop signals competing for the 1536-byte cap.
4. **P3**: decide whether `teamId` on `WakeCardRequest` becomes a real filter or gets stripped.
5. **P3**: lex-tiebreaker on `pickCurrentWorkItem` for equal timestamps.

PR #424, opened 2026-05-04
P0-5: orc End-of-Turn Delivery Verification (pre-yield self-check)

stevehuang0115:
## Summary

- Add a 4-step self-check the orchestrator runs **before yielding the turn**: verify every inbound `[CHAT:slack-...]` message produced at least one `reply-slack/execute.sh` call.
- Catches silent-fail Slack delivery — the bug Steve has had to ping "?" to surface.
- Net-zero offset: cut a redundant one-line `**CRITICAL**:` stub that restated the #1 notification rule.

## Spec reference

`.crewly/specs/2026-05-03-agent-improvement-p0-execution.md` — **Fix P0-5** (lines 196–220).

Spec gives the exact prompt text; this PR places it as a top-level `## End-of-Turn Delivery Verification` section, positioned right after the existing `## CRITICAL: Notification Protocol — ALWAYS RESPOND TO THE USER` section so orc reads them as one coherent pre-yield checklist.

## What was added

```markdown
## End-of-Turn Delivery Verification

Before yielding the turn:
1. Did I receive any `[CHAT:slack-...]` messages this turn?
2. For EACH such message, did I make at least one Bash call to `reply-slack/execute.sh`?
3. If no, the response was NOT delivered — `[NOTIFY]` alone is not sufficient.
4. If the answer to (2) is "no," call `reply-slack` now BEFORE yielding the turn.

This is a hard pre-yield check. Do not yield if any Slack message is unanswered.
```

## What was cut (net-zero offset)

The redundant one-liner from the Smart Event Notification Protocol section:

```markdown
**CRITICAL**: Plain text output (without markers) goes nowhere — the user won't see it in Chat or Slack. You MUST use `[NOTIFY]` markers for Chat UI updates and `reply-slack` skill for Slack messages.
```

**Why this stub:** it was a one-line restatement of two existing rules:
- Line 263 (the **#1 rule** of the Notification Protocol): "Every `[CHAT:...]` message MUST produce at least one `[NOTIFY]` response."
- Line 784 (the CRITICAL ANTI-PATTERN): "Receiving a `[CHAT:...]` message, then running 3-5 bash scripts without ever outputting a `[NOTIFY]`."

The new End-of-Turn Delivery Verification section subsumes the same point with concrete steps, so the stub is true semantic redundancy — not a unique rule.

## Char budget proof

| Metric | Before | After | Δ |
|---|---|---|---|
| Chars | 81,460 | 81,715 | **+255 (+0.31%)** |
| Lines | 1,584 | 1,592 | +8 |
| Diff | — | — | +10 / -2 |

**Substantively net-zero** on an 81K-char prompt. New section adds ~510 chars (4-step protocol with rationale); cut stub freed ~255 chars. The +255 delta is the price of a structured 4-step check vs. a one-liner.

## Manual reasoning trace — does this catch yesterday's silent-fail?

Pre-yield check applied to the failure mode the spec calls out (orc writes Slack response in terminal text but never calls `reply-slack`):

1. **Step 1** — "Did I receive any `[CHAT:slack-...]` messages this turn?" → **YES** (the user pinged orc on Slack)
2. **Step 2** — "For EACH such message, did I make at least one Bash call to `reply-slack/execute.sh`?" → **NO** (terminal text only)
3. **Step 3** — "If no, the response was NOT delivered — `[NOTIFY]` alone is not sufficient." (orc recognizes the gap)
4. **Step 4** — "Call `reply-slack` now BEFORE yielding the turn." (orc self-corrects)

→ **Caught.** Owner does not need to ping "?" because orc self-corrects before yielding. ✅

## Files changed

- `config/roles/orchestrator/prompt.md` (+10 / -2)

Single file. Prompt-only change. No backend / messaging / skills code touched.

## Risk

- **Prompt budget:** +0.31% on an 81K prompt — negligible.
- **Behavior regression:** the cut stub is true semantic redundancy (verified via grep — no other prompt file references this exact phrasing); removing it does not weaken any unique behavior.
- **In-flight collisions:** branched from `origin/main` HEAD `ad817c3c` (post #418 Periodic Progress Check-In). No in-flight PRs touch the same lines.
- **Self-evolution loop:** confirmed — `self-evolution.md` does not reference the cut stub or the inserted lines.

## Test plan

- [x] `git diff --stat` shows single-file change (+10/-2)
- [x] `wc -c` baseline + post-edit captured (net +255 / +0.31%)
- [x] Manual reasoning trace: confirms the rule catches the silent-fail pattern the spec calls out
- [x] Self-evolution.md skim: no overlap with edited regions
- [x] grep for the cut phrasing across all orc prompt files: only one occurrence existed (the one I cut)
- [ ] Post-merge real-world verification: zero silent-fail Slack delivery incidents in week 1 (per spec success metric)

## Out of scope

- **P0-6** (Owner-Facing Communication Standard) — same spec, separate fix, separate PR.
- Any non-orchestrator prompt changes.

### DIFF
--- a/config/roles/orchestrator/prompt.md
+++ b/config/roles/orchestrator/prompt.md
@@ -421,6 +421,16 @@ bash {{ORCHESTRATOR_SKILLS_PATH}}/reply-slack/execute.sh '{"channelId":"D0AC7NF5
 5. **Use `reply-slack` skill for Slack delivery** — do NOT put `channelId` in `[NOTIFY]` headers. Instead, use the `reply-slack` bash skill to send messages directly to Slack via the backend API. This avoids PTY terminal artifacts that garble Slack messages. Use `[NOTIFY]` (with `conversationId`) for Chat UI only.
 6. **No JSON escaping needed** — write markdown naturally in the body after `---`
 
+## End-of-Turn Delivery Verification
+
+Before yielding the turn:
+1. Did I receive any `[CHAT:slack-...]` messages this turn?
+2. For EACH such message, did I make at least one Bash call to `reply-slack/execute.sh`?
+3. If no, the response was NOT delivered — `[NOTIFY]` alone is not sufficient.
+4. If the answer to (2) is "no," call `reply-slack` now BEFORE yielding the turn.
+
+This is a hard pre-yield check. Do not yield if any Slack message is unanswered.
+
 ## Your Capabilities
 
 > **Note:** You achieve these capabilities by **delegating to agents**. Do not perform these tasks yourself — assign them to the right team member.
@@ -755,8 +765,6 @@ Then for Slack:
 ```bash
 bash {{ORCHESTRATOR_SKILLS_PATH}}/reply-slack/execute.sh '{"channelId":"C0123","text":"*Emily (5-min check)*\nActively working on visa.careerengine.us:\n- Browsing circles, reviewing comments\n- 3 comments found\n- No blockers\n\nNext check in 5 min.","threadTs":"170743.001"}'
 ```
-
-**CRITICAL**: Plain text output (without markers) goes nowhere — the user won't see it in Chat or Slack. You MUST use `[NOTIFY]` markers for Chat UI updates and `reply-slack` skill for Slack messages.
 
 ### Proactive Behaviors You Should Always Do
 


stevehuang0115 (commented):

Fast-tier review of #424. Pure markdown, +10/-2, single-file. Net: ship it.

---

### What this PR does

Inserts a new top-level `## End-of-Turn Delivery Verification` section right after the Notification Protocol's `### Important Rules` block. Codifies a 4-step pre-yield self-check: did I receive a `[CHAT:slack-...]` this turn → for each one, did I make at least one `reply-slack/execute.sh` Bash call → if no, the response was NOT delivered → if no, call reply-slack now BEFORE yielding.

Closes the silent-fail Slack-delivery bug Steve has had to ping "?" to surface — orc writes a Slack response in terminal text but never invokes the skill, the user sees nothing, the orc thinks it answered.

Net-zero offset: cuts a redundant `**CRITICAL**: Plain text output (without markers) goes nowhere...` stub from the Smart Event Notification Protocol section.

---

### Reality-check on the redundancy claim (the reviewer-judgment-dependent part)

The PR description claims the cut stub was a restatement of "Line 263 (the #1 rule of the Notification Protocol)" and "Line 784 (the CRITICAL ANTI-PATTERN)". I cross-checked.

**The redundancy claim holds, but the line citations in the PR description are off:**

- Line 263 in the pre-PR file is the "Pre-check validation" block of the recurring-check cycle, NOT the #1 Notification Protocol rule.
- Line 784 in the pre-PR file is the V3 Intelligent Decomposition Pipeline, NOT the CRITICAL ANTI-PATTERN.

The actual subsumption sources are:

1. **`### Important Rules` #1** (post-edit line 421-422): *"NEVER let a chat message go unanswered — every `[CHAT:...]` MUST get a `[NOTIFY]`."* Covers the chat-must-respond half.
2. **The surviving `**RULE:**` line** (post-edit line 777, pre-edit line 733, ~50 lines after the cut site): *"Every proactive update MUST use `[NOTIFY]` markers with `conversationId` for Chat UI AND `reply-slack` skill for Slack. Plain text output is invisible to the user — it only appears in the terminal log."* Covers the same NOTIFY+reply-slack mandate the cut stub stated, in stronger phrasing.

So the stub really is true semantic redundancy — but the redundancy lives next door (line 777 in post-edit), not where the PR description points. The cut is safe; the description's line numbers just need to be re-anchored if anyone audits the trail. Minor doc hygiene, not a merge blocker.

---

### Section placement coherence

The new section lands between the `### Important Rules` of the Slack Reply Protocol (lines 415-422) and `## Your Capabilities` (line 434). That's the right neighborhood:

- The preceding `### Important Rules` establishes *what to do* (every CHAT must get a NOTIFY + use reply-slack for Slack delivery)
- The new `## End-of-Turn Delivery Verification` establishes *how to verify you actually did it* before yielding

Adjacent placement reads as one coherent pre-yield flow, which matches the PR description's stated intent ("a coherent pre-yield checklist"). ✅

---

### Diff line-by-line

`config/roles/orchestrator/prompt.md:424-432` (new section):

- ✅ **Step 1** (`Did I receive any [CHAT:slack-...] messages this turn?`) — clear yes/no gate. Slack-prefixed pattern is the right scope (Chat-UI `[CHAT:conversationId]` paths render via `[NOTIFY]` automatically; the silent-fail mode is Slack-specific because Slack requires the explicit skill call).
- ✅ **Step 2** (`For EACH such message, did I make at least one Bash call to reply-slack/execute.sh?`) — uses "Bash" as the tool-type cue and the exact skill path. Counts skill calls per message, not per turn — defends against the multi-message-collapsed-to-one-reply failure mode.
- ✅ **Step 3** (`If no, the response was NOT delivered — [NOTIFY] alone is not sufficient`) — sharp clarification, surfaces the precise misconception the silent-fail bug is built on (orc thinks `[NOTIFY]` covers Slack; it doesn't).
- ✅ **Step 4** (`call reply-slack now BEFORE yielding the turn`) — actionable directive.
- ✅ **Closing line** (`This is a hard pre-yield check. Do not yield if any Slack message is unanswered.`) — unambiguous; "hard" closes any wiggle room about the check being aspirational.

`config/roles/orchestrator/prompt.md:759` (the cut, now line 769 in post-edit):

- ✅ Cuts only the standalone `**CRITICAL**` paragraph that lived between the example bash block and the `### Proactive Behaviors You Should Always Do` heading. Surrounding structure intact; section flow reads cleanly without the stub.

---

### Self-evolution overlap check

Grep'd `self-evolution.md` for both the cut-stub phrasing and the new-section anchors (`End-of-Turn Delivery`, `reply-slack/execute`). **Zero hits.** No self-evolution overlap; no risk of an orc auto-amend reverting either side of this change.

---

### Char budget claim

PR claims 81,460 → 81,715 chars (+255, +0.31%). I measured the head file: **81,715 chars exact match** on the "after" number. Char delta is real and small.

PR claims 1,584 → 1,592 lines (+8). My measurement of the head file: **1,628 lines.** That's a 36-line gap on the absolute number — but the +8 *delta* matches the +10/-2 diff stat, and the absolute mismatch is consistent with Quinn baselining against a pre-#418 snapshot (#418 added +36 lines to the same file). PR description could be footnoted, but the delta arithmetic is correct. Non-blocker.

Either way, +0.31% on an 81K prompt is well under the gilding threshold.

---

### In-flight collision check

PR description states it branched from post-#418 main. `gh pr view 424` confirms `mergeable: MERGEABLE`. The new section lands at line 424; #418's Periodic Progress Check-In landed at line 30-65. No overlap with #418, and no other in-flight prompt PRs touch this file.

---

### Behavioral judgment — does the 4-step actually catch the bug?

Trace the failure mode (orc receives `[CHAT:slack-...]`, writes response in terminal text, never calls `reply-slack`, yields):

1. Step 1: `Did I receive any [CHAT:slack-...] messages this turn?` → **YES**
2. Step 2: `For EACH such message, did I make at least one Bash call to reply-slack/execute.sh?` → **NO** (terminal text only)
3. Step 3: `If no, the response was NOT delivered — [NOTIFY] alone is not sufficient.` → orc recognizes the gap (this is the lesson the bug taught us)
4. Step 4: `call reply-slack now BEFORE yielding the turn` → orc self-corrects

✅ Caught. Steve doesn't need to ping "?".

The 4-step is a Boolean check on tool-call presence, which is something an LLM can reliably do by inspecting its own turn's tool-use record. This is a strong design choice — much more reliable than a "remember to send Slack" prose nudge.

---

1. **PR description line citations are stale.** Description points at lines 263 / 784 for the subsumption sources; actual sources are line 421-422 (Important Rules #1) and line 777 (the surviving RULE). Underlying redundancy claim still holds. Worth a post-merge edit to the PR body so the audit trail is accurate, but not a merge blocker.
2. **Line-count baseline appears stale.** PR table shows 1,584→1,592 but actual file is 1,628 lines (consistent with Quinn baselining pre-#418). +8 delta is correct; absolute numbers are off. Non-blocker.
3. **Step 1 scopes to `[CHAT:slack-...]` only.** Chat UI `[CHAT:<conversationId>]` (without `slack-` prefix) isn't covered by the verification. This is **intentionally correct** — the silent-fail bug is Slack-specific because Chat UI gets `[NOTIFY]` rendering automatically. Worth a one-line clarification in a follow-up to lock the scope reasoning into the prompt itself, but not a blocker.
4. **Threading not specified.** If the orc is replying to a threaded Slack message, step 4 doesn't say "preserve the threadTs". Existing examples earlier in the prompt (line 769ish) already show `threadTs` usage in `reply-slack` calls, so the orc has the pattern, but a one-word reminder ("call `reply-slack` with the original `threadTs`") would tighten the contract. P3.
5. **Verification doesn't enforce content of the reply.** Step 4 catches "no reply at all" but not "trivial reply that doesn't answer the question". That's the right scope for this fix (delivery-layer verification, not content-layer). Content-layer is presumably P0-6 (Owner-Facing Communication Standard, called out as out-of-scope).
6. **Surviving line-777 RULE could swallow the new section** in a future cleanup pass if someone misreads them as duplicates. They're not — line 777 is a *do-this* rule; the new section is a *verify-you-did* check. Worth a one-line cross-reference between them ("Verify per the End-of-Turn check below") to prevent future confusion. P3.

---

### Recommendation

**Merge as-is.** Tight markdown, additive only on net char count, every cross-reference verified, redundancy claim holds (despite line-citation drift in the description), placement is coherent, the 4-step actually catches the failure mode it's designed for, and self-evolution.md has no overlap. Owner-veto (Steve) keeps cadence and tone under owner control via separate mechanisms.

**Suggested follow-ups (none blocking, all P2-P3):**
1. **P2**: amend PR description to point at the *real* subsumption sources (post-edit line 421-422 + line 777) for audit-trail accuracy.
2. **P3**: add a one-line scope clarification to step 1 ("Chat UI `[CHAT:<conversationId>]` messages are covered by `[NOTIFY]` rendering; this check is Slack-specific because Slack requires the skill call.").
3. **P3**: add `threadTs` reminder to step 4 to defend against accidentally posting to channel root instead of the original thread.
4. **P3**: cross-reference the new section from the surviving line-777 RULE so future readers see they're a do/verify pair, not duplicates.

Welcome aboard for round two, Quinn — clean prompt-only fix with a sharp behavioral target. The 4-step structure (yes/no gate → enforcement directive) is a good pattern for any future delivery-layer self-checks.

PR #425, opened 2026-05-04
fix(frontend): complete DeepSeek Phase 1 — wire deepseek into ApiKeyProvider + Settings UI

stevehuang0115:
## Summary

Closes the DeepSeek Phase 1 frontend gap Sam's ralph-loop tick flagged. Backend already accepts `deepseek` as an `ApiKeyProvider` (see `backend/src/types/settings.types.ts:124,133` + `model-manager.ts`), but the frontend type union and Settings UI were still `gemini | anthropic | openai`-only — so end-users could **not** save a `DEEPSEEK_API_KEY` through Settings UI. They had to `export DEEPSEEK_API_KEY=...` manually. Phase 1 didn't actually reach the user.

This PR mirrors the backend types exactly and renders the DeepSeek key field + runtime override row in `ApiKeysTab`.

## What changed

- **`frontend/src/types/settings.types.ts`**
  - `ApiKeyProvider` union extended: `'gemini' | 'anthropic' | 'openai' | 'deepseek'`
  - `API_KEY_PROVIDERS` array gains `'deepseek'` (drives UI iteration)
  - `ApiKeysSettings.global` / `.runtimeOverrides[*]` / `.skillOverrides[*]` interfaces all gain optional `deepseek` slots
  - Provider order now matches backend (`gemini, anthropic, openai, deepseek`) so UI rows render in the same sequence as backend lookups
- **`frontend/src/components/Settings/ApiKeysTab.tsx`**
  - `PROVIDER_DISPLAY_NAMES`: `+ deepseek: 'DeepSeek'`
  - `PROVIDER_ENV_HINTS`: `+ deepseek: 'DEEPSEEK_API_KEY'`

## Tests

- **`settings.types.test.ts`** (+4): asserts `API_KEY_PROVIDERS` contains all 4 providers, exact length === 4, exact order matches backend, compile-time assignment of `'deepseek' as ApiKeyProvider`
- **`ApiKeysTab.test.tsx`** (+2 new, +4 updated):
  - NEW: input acceptance on DeepSeek field + saved-key rendering
  - UPDATED: provider-row count `3 → 4`, test-button count `3 → 4`, env-var hint count `3 → 4`, "Not set" indicator count `3 → 4`
- **All 25 affected tests pass** (`vitest run` on the two files)
- **Frontend TS clean** on changed files (preexisting unrelated chat-ui errors are not introduced by this PR)
- **Vite build clean**

## Manual smoke (deferred to staging)

- [ ] Open Settings → API Keys → enter DeepSeek key → Save → confirm `apiKeys.global.deepseek` persists
- [ ] Spawn a `crewly-agent` runtime with a DeepSeek model → confirm `model-manager.getApiKey('deepseek')` returns the saved key (already wired in backend)

## Why this matters

Backend type-drift was already done correctly here — the failure mode was the **frontend mirror missing**. Same lesson as PR #421's dead-path bug: in multi-layer systems, type-union drift between layers is a silent killer that only surfaces when a real user tries to use the feature. Worth a watcher / lint rule on next iteration.

## Coordination

Phase 1 ship-blocking work only; no shared module surface with Sam's Phase 2 I2 reasoning_content extraction (different files, different concerns). Phase 2 should proceed once this lands.

## Refs

- Backend canonical types: `backend/src/types/settings.types.ts:124,133,139,143`
- Backend runtime wiring: `backend/src/services/agent/crewly-agent/model-manager.ts:107-118` (DeepSeek via OpenAI-compat at `https://api.deepseek.com/v1`)

### DIFF
--- a/frontend/src/components/Settings/ApiKeysTab.test.tsx
+++ b/frontend/src/components/Settings/ApiKeysTab.test.tsx
@@ -67,6 +67,7 @@ describe('ApiKeysTab', () => {
     expect(screen.getByText('Google Gemini')).toBeInTheDocument();
     expect(screen.getByText('Anthropic')).toBeInTheDocument();
     expect(screen.getByText('OpenAI')).toBeInTheDocument();
+    expect(screen.getByText('DeepSeek')).toBeInTheDocument();
   });
 
   it('should render runtime override sections', () => {
@@ -108,7 +109,7 @@ describe('ApiKeysTab', () => {
   it('should render test buttons for each provider', () => {
     render(<ApiKeysTab />);
     const testButtons = screen.getAllByText('Test');
-    expect(testButtons.length).toBe(3); // One per global provider
+    expect(testButtons.length).toBe(4); // One per global provider (gemini, anthropic, openai, deepseek)
   });
 
   it('should show env var hints', () => {
@@ -116,6 +117,7 @@ describe('ApiKeysTab', () => {
     expect(screen.getByText('GOOGLE_GENERATIVE_AI_API_KEY')).toBeInTheDocument();
     expect(screen.getByText('ANTHROPIC_API_KEY')).toBeInTheDocument();
     expect(screen.getByText('OPENAI_API_KEY')).toBeInTheDocument();
+    expect(screen.getByText('DEEPSEEK_API_KEY')).toBeInTheDocument();
   });
 
   it('should render with existing global keys', () => {
@@ -142,6 +144,38 @@ describe('ApiKeysTab', () => {
   it('should show "Not set" for unconfigured keys', () => {
     render(<ApiKeysTab />);
     const notSetElements = screen.getAllByText('Not set');
-    expect(notSetElements.length).toBe(3);
+    expect(notSetElements.length).toBe(4);
+  });
+
+  it('should accept input on the DeepSeek global key field', () => {
+    render(<ApiKeysTab />);
+    const input = screen.getByPlaceholderText('Enter DEEPSEEK_API_KEY') as HTMLInputElement;
+    expect(input).toBeInTheDocument();
+    fireEvent.change(input, { target: { value: 'sk-deepseek-test-1234' } });
+    expect(input.value).toBe('sk-deepseek-test-1234');
+  });
+
+  it('should render saved DeepSeek global key', async () => {
+    mockUseSettings.mockReturnValue({
+      settings: {
+        general: { defaultRuntime: 'claude-code' as const },
+        chat: {},
+        skills: {},
+        apiKeys: {
+          global: { deepseek: '••••••••wxyz' },
+          runtimeOverrides: {},
+          skillOverrides: {},
+        },
+      },
+      updateSettings: mockUpdateSettings,
+      isLoading: false,
+      error: null,
+    });
+    render(<ApiKeysTab />);
+    // Three providers (gemini/anthropic/openai) report "Not set"; deepseek reports "Saved".
+    await waitFor(() => {
+      expect(screen.getByText('Saved')).toBeInTheDocument();
+      expect(screen.getAllByText('Not set').length).toBe(3);
+    });
   });
 });
--- a/frontend/src/components/Settings/ApiKeysTab.tsx
+++ b/frontend/src/components/Settings/ApiKeysTab.tsx
@@ -32,6 +32,7 @@ const PROVIDER_DISPLAY_NAMES: Record<ApiKeyProvider, string> = {
   gemini: 'Google Gemini',
   anthropic: 'Anthropic',
   openai: 'OpenAI',
+  deepseek: 'DeepSeek',
 };
 
 /**
@@ -41,6 +42,7 @@ const PROVIDER_ENV_HINTS: Record<ApiKeyProvider, string> = {
   gemini: 'GOOGLE_GENERATIVE_AI_API_KEY',
   anthropic: 'ANTHROPIC_API_KEY',
   openai: 'OPENAI_API_KEY',
+  deepseek: 'DEEPSEEK_API_KEY',
 };
 
 type SaveStatus = 'idle' | 'saving' | 'saved' | 'error';
--- a/frontend/src/types/settings.types.test.ts
+++ b/frontend/src/types/settings.types.test.ts
@@ -8,6 +8,8 @@ import { describe, it, expect } from 'vitest';
 import {
   AI_RUNTIMES,
   AI_RUNTIME_DISPLAY_NAMES,
+  API_KEY_PROVIDERS,
+  ApiKeyProvider,
   getAIRuntimeDisplayName,
   isValidAIRuntime,
 } from './settings.types';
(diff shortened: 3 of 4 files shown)


stevehuang0115 (commented):

Standard review of #425 against Steve's flagged check (frontend ↔ backend type mirror, since drift was the previous Phase 1 ship-blocker root cause). Net: ship it.

---

### Frontend ↔ backend mirror — the Steve-flagged check

This was the explicit verify-this from the dispatch ("frontend 跟 backend types 完全 mirror — 这条 mismatch drift 是上次 Phase 1 ship 漏掉的根因").

**Confirmed: PASS — exact mirror across all four type surfaces.**

| Surface | Backend (`backend/src/types/settings.types.ts`) | Frontend (this PR) | Match? |
|---|---|---|---|
| `ApiKeyProvider` union | `'gemini' \| 'anthropic' \| 'openai' \| 'deepseek'` (line 124) | `'gemini' \| 'anthropic' \| 'openai' \| 'deepseek'` | ✅ exact |
| `API_KEY_PROVIDERS` const array | `['gemini', 'anthropic', 'openai', 'deepseek']` (lines 129-134) | `['gemini', 'anthropic', 'openai', 'deepseek']` | ✅ exact (incl. order) |
| `ApiKeysSettings.global.deepseek` | `deepseek?: string` (line 168) | `deepseek?: string` | ✅ exact |
| `ApiKeysSettings.runtimeOverrides[*].deepseek` | `deepseek?: ApiKeyConfig` (line 177) | `deepseek?: ApiKeyConfig` | ✅ exact |
| `ApiKeysSettings.skillOverrides[*].deepseek` | `deepseek?: ApiKeyConfig` (line 187) | `deepseek?: ApiKeyConfig` | ✅ exact |

**Order is part of the contract** (UI iterates the array to render rows in the same sequence as backend lookups). The order test (`expect([...API_KEY_PROVIDERS]).toEqual(['gemini', 'anthropic', 'openai', 'deepseek'])`) pins this against future shuffle.

---

### TypeScript-enforced exhaustiveness — the structural safety net

Because `PROVIDER_DISPLAY_NAMES` and `PROVIDER_ENV_HINTS` are typed as `Record<ApiKeyProvider, string>`, **tsc will fail** if either map is missing a key for any union member. So the PR's `+ deepseek: 'DeepSeek'` and `+ deepseek: 'DEEPSEEK_API_KEY'` lines aren't optional polish — they're required for tsc clean. **tsc passing is itself proof of completeness.**

This is the right structural safety: extending the union without updating the maps is a build error, not a silent runtime gap. Future provider additions will fail loudly at the same boundary.

---

### Render mechanics — does adding 4 lines actually render the DeepSeek row everywhere it should?

`ApiKeysTab.tsx` has **two render sites** that iterate the providers:

1. **Line 249** — Global API Keys section: `{API_KEY_PROVIDERS.map(provider => ...)}` → renders one row per provider with `PROVIDER_DISPLAY_NAMES[provider]` label + `PROVIDER_ENV_HINTS[provider]` placeholder + Test button + "Saved" / "Not set" indicator.
2. **Line 331** — Runtime Overrides section (per-runtime, expandable): `{API_KEY_PROVIDERS.map(provider => ...)}` → renders one override row per provider per runtime.

Both render sites use the same `API_KEY_PROVIDERS.map(...)`. Extending the array + the two `Record<ApiKeyProvider, string>` maps is mathematically sufficient to render the DeepSeek row in **both** sections automatically. ✅

The test count updates (`3 → 4` for provider rows, test buttons, env hints, "Not set" indicators) all line up with this mechanic.

**Skill Overrides UI section** (line 62 initializes `skillOverrides: {}` in state) is **not currently rendered** in the component — the type field on `ApiKeysSettings.skillOverrides[*]` is forward-compat only. Adding `deepseek?: ApiKeyConfig` to the type is harmless if the UI doesn't iterate; correct if/when the Skill Overrides UI eventually lands. ✅

---

### Backend wiring — already in place, this PR closes the loop

Per the PR description's `model-manager.ts:107-118` reference, verified the backend wiring:

- **Line 33**: `DEEPSEEK_BASE_URL = 'https://api.deepseek.com/v1'`
- **Lines 107-118**: `case 'deepseek':` constructs `deepseekProvider` via `createOpenAI({ baseURL: DEEPSEEK_BASE_URL, apiKey: process.env.DEEPSEEK_API_KEY })`. Comment explicitly notes the `.chat(modelId)` requirement (DeepSeek implements `/chat/completions` only, not `/responses`).
- **Lines 147-159**: resolves `deepseek` via `settingsService.getApiKey('deepseek', context)` — settings-saved key piped to the runtime.
- **Lines 206-207**: `case 'deepseek': process.env.DEEPSEEK_API_KEY = key;` — writes the resolved key to env so the OpenAI-compat SDK picks it up.

End-to-end chain: **Settings UI → settingsService → `process.env.DEEPSEEK_API_KEY` → model-manager → DeepSeek API**. Backend was ready; the user just had no UI surface to reach it before this PR.

---

### Test plan adequacy

- **`settings.types.test.ts` +4** — content (all 4 providers present), exact length (`toHaveLength(4)`), exact order match against backend, compile-time `'deepseek' as ApiKeyProvider` assignment. The order assertion is the regression pin against future shuffle.
- **`ApiKeysTab.test.tsx` +2 new + 4 updated**:
  - **NEW `should accept input on the DeepSeek global key field`** — finds input by `placeholder='Enter DEEPSEEK_API_KEY'`, fires change event, asserts value. Proves the iteration through `PROVIDER_ENV_HINTS` correctly produces the placeholder.
  - **NEW `should render saved DeepSeek global key`** — sets `apiKeys.global.deepseek = '••••••••wxyz'` (no other keys set), asserts exactly one "Saved" + exactly 3 "Not set". Sharp test — proves the deepseek slot reads from `apiKeys.global.deepseek` specifically (no slot-confusion).
  - **UPDATED count assertions** (`3 → 4` for provider rows, test buttons, env hints, "Not set" indicators) — all internally consistent.
- **25/25 affected tests pass**, vite build clean, frontend tsc clean on changed files.

The "render saved DeepSeek global key" test is exactly the regression pin you want for a slot-mapping change. ✅

---

1. **No cross-package drift-prevention test.** Today the frontend `API_KEY_PROVIDERS` and backend `API_KEY_PROVIDERS` are two literal-equal arrays in two different packages, kept in sync by reviewer attention. The PR description correctly calls this out: *"Worth a watcher / lint rule on next iteration."* A unit test that imports both arrays and asserts equality would catch the next drift automatically. Possible shapes: (a) backend integration test that imports the frontend types via relative path; (b) extract `ApiKeyProvider` + `API_KEY_PROVIDERS` to a shared `types/` package both consume; (c) lightweight CI script that diffs the two declarations. **P1 follow-up** — same structural risk that produced the original drift.

2. **`PROVIDER_ENV_HINTS` shows one env var per provider; backend `API_KEY_ENV_VARS` allows an array.** Backend `gemini` accepts both `GOOGLE_GENERATIVE_AI_API_KEY` and `GEMINI_API_KEY` (line 140). Frontend hint shows only the first. Minor UX: a user who exported `GEMINI_API_KEY` would see "Not set" in the UI while the backend would actually find the key. Out-of-scope for this PR (pre-existing behavior, not introduced by deepseek). **P3** — could change `PROVIDER_ENV_HINTS` to import `API_KEY_ENV_VARS[provider][0]` from the shared types if the cross-package refactor (P1 above) lands.

3. **`PROVIDER_DISPLAY_NAMES` + `PROVIDER_ENV_HINTS` live in `ApiKeysTab.tsx` only.** They're used here and only here. If a second component ever needs to render a provider name (e.g., a runtime selector), the two will drift. **P3** — colocate with `settings.types.ts` once a second consumer materializes; premature now.

4. **Skill Overrides UI gap (forward-compat).** `ApiKeysSettings.skillOverrides[*].deepseek?: ApiKeyConfig` type field is defined but no UI section iterates skill overrides today (line 62 just initializes `skillOverrides: {}`). The new `deepseek` slot in this type is harmless if/when the Skill Overrides UI lands. Mention only because reviewers might wonder why there's no skill-override test for deepseek.

5. **Backend `DEEPSEEK_BASE_URL` is hardcoded** (`backend/src/services/agent/crewly-agent/model-manager.ts:33`). For DeepSeek mainland-China deployments that route to a different endpoint, this is a hard-code. Out-of-scope for this PR but worth flagging as **P2 / pro-tier follow-up** to allow per-runtime baseURL override (probably belongs in the `ApiKeyConfig` shape eventually).

---

### Recommendation

**Merge as-is.** Frontend mirror is exact across all four type surfaces (union, const array, three nested ApiKeysSettings slots) and the order matches backend. Render mechanics are TypeScript-enforced exhaustive via the two `Record<ApiKeyProvider, string>` maps — tsc would fail loudly on incomplete coverage. Both render sites (Global + Runtime Overrides) iterate `API_KEY_PROVIDERS` so adding the array entry + two map entries automatically renders the DeepSeek row everywhere it should appear. End-to-end chain (UI → settings → env → model-manager) closes with this PR.

1. **P1**: add a cross-package drift-prevention test (or extract `ApiKeyProvider` + `API_KEY_PROVIDERS` to a shared types module both packages consume) — same structural risk that produced the original Phase 1 ship-blocker.
2. **P2 / pro-tier**: allow per-runtime baseURL override on `ApiKeyConfig` for DeepSeek mainland-China deployments and other OpenAI-compat endpoints.
3. **P3**: change `PROVIDER_ENV_HINTS` to read `API_KEY_ENV_VARS[provider][0]` from shared types once P1 lands, so frontend hints automatically pick up the canonical env var name.
4. **P3**: colocate `PROVIDER_DISPLAY_NAMES` + `PROVIDER_ENV_HINTS` with `settings.types.ts` if/when a second consumer materializes.

Same lesson the PR description names ("type-union drift between layers is a silent killer") is correct — and this PR is the right closing move for Phase 1 user-reach. Phase 2 (Sam's I2 reasoning_content extraction) can proceed independently.

PR #447, opened 2026-05-05
fix(compact): SI A+B follow-up — close 3 gaps from PR #444 audit

stevehuang0115:
## Why

PR #444 flipped the `GeneralSettings.enableProactiveCompact` default from `true` → `false`, but Arch's compact-code audit (`/tmp/compact_audit_verdict.md` — local file, distributed below) showed **PR #444 IS NECESSARY BUT INSUFFICIENT**. Three gaps left the proactive `/compact` trigger firing for every existing user in production:

1. **Persisted settings overrode the new default** — `mergeSettings(defaults, loaded)` overlays loaded values on top, so an existing `~/.crewly/settings.json` with `enableProactiveCompact: true` (Steve's, and every existing user's) wins over PR #444's new `false` default. PR #444 only affected fresh installs.
2. **Initial cached `proactiveCompactEnabled = true` + fallback `?? true` were fail-OPEN** — the very first `performCheck()` after process start ran with proactive ENABLED regardless of settings, and any settings-read failure kept proactive ON.
3. **No migration** — `migrateSettings()` had no code path to flip the persisted value for existing users.

Per Steve's user-position decision (2026-05-04): "Claude Code self-compacts internally, external compact is unnecessary." PR #444 alone could not enforce that; this PR does.

### §A: Migration + cache fail-safe

- **A.1** — `SettingsService.migrateSettings()` adds an idempotent one-time flip: if persisted `enableProactiveCompact === true` and the new `_proactiveCompactMigrated` marker is absent, flip to `false` and stamp the marker. Users who deliberately re-enable via Settings UI are NOT re-flipped on subsequent loads.
- **A.2** — `context-window-monitor.service.ts` initial cached `proactiveCompactEnabled` flipped from `true` → `false`, and the settings-read fallback flipped from `?? true` → `?? false`. Fail-safe **CLOSED** — when uncertain, the wrapper does NOT compact.

### §B: Threshold-compact gating

- New `GeneralSettings.enableThresholdCompact: boolean` (default `false`) gates the RED (≥85%) and CRITICAL (≥95%) compact paths.
  - When closed: RED pre-compaction flush + `/compact` is skipped; CRITICAL `/compact` attempt is skipped.
  - **Auto-recovery still runs at CRITICAL** when `AUTO_RECOVERY_ENABLED` — the safety net is independent of the threshold-compact gate.
  - The `agent:context_critical` event is still published regardless of gate (informational).

## Test plan

All 1:1 source-to-test per CLAUDE.md.

- [x] `settings.types.test.ts` — 2 new tests (default + validation), 3 inline `GeneralSettings` literals updated. **85/85 pass**
- [x] `settings.service.test.ts` — 4 new tests covering: flip on first load, idempotent re-enable preserved, no-op when already `false`, no-op when missing. **58/58 pass**
- [x] `context-window-monitor.service.test.ts` — 11 new tests in `compact gating (Option A+B)` describe block:
  - initial cached defaults `false` for both gates
  - RED-level compact BLOCKED when `thresholdCompactEnabled=false`
  - RED-level compact FIRES when `thresholdCompactEnabled=true`
  - CRITICAL-level compact attempt BLOCKED when `thresholdCompactEnabled=false`
  - CRITICAL still publishes `agent:context_critical` regardless of gate
  - both refresh paths fall back to `false` on undefined settings
  - both refresh paths read `true` when persisted
  - threshold refresh keeps last-known on read failure
  
  Outer `beforeEach` re-opens both gates for behavior-style tests. **101/101 pass** (existing 90 + new 11).

## Rollback path

- Users opt back into proactive compact via Settings UI. The migration marker prevents re-flip.
- Users opt into threshold compact by setting `enableThresholdCompact: true` in `~/.crewly/settings.json`.

## Coordination

Surfaced for Arch deep-review. ORC + Steve aware via TL coord. Merge after Arch ✓.

### DIFF
--- a/backend/src/services/agent/context-window-monitor.service.test.ts
+++ b/backend/src/services/agent/context-window-monitor.service.test.ts
@@ -171,6 +171,18 @@ describe('ContextWindowMonitorService', () => {
 		jest.clearAllMocks();
 		jest.useFakeTimers();
 		ContextWindowMonitorService.resetInstance();
+
+		// Per spec 2026-05-05-compact-fix-AB-followup §A.2 + §B the production
+		// defaults for both compact gates flipped to `false` (fail-safe CLOSED).
+		// Existing behavior-tests in this suite were written when the gates were
+		// effectively `true` and they assert that compact fires at the threshold.
+		// Re-open both gates here so those behavior assertions still hold without
+		// rewriting every test. The new `compact gating (Option A+B)` describe
+		// block intentionally OVERRIDES these to `false` to verify the gating
+		// contract (defaults closed, RED/CRITICAL paths blocked when off, etc).
+		const service = ContextWindowMonitorService.getInstance();
+		(service as unknown as { proactiveCompactEnabled: boolean; thresholdCompactEnabled: boolean }).proactiveCompactEnabled = true;
+		(service as unknown as { proactiveCompactEnabled: boolean; thresholdCompactEnabled: boolean }).thresholdCompactEnabled = true;
 	});
 
 	afterEach(() => {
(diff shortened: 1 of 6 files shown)


stevehuang0115:
Pushed obs follow-up commit `befc0738` addressing Arch's 3 non-blocking observations:

1. ✅ Direct `_proactiveCompactMigrated: true` cast-and-assert in the flip test (settings.service.test.ts)
2. ✅ JSDoc note in `migrateSettings()` clarifying marker is in-memory-until-saveSettings (settings.service.ts)
3. ✅ Negative gating test for proactive path with `proactiveCompactEnabled=false` (context-window-monitor.service.test.ts)

Test count 244 → 245 (3 suites green: 85 + 59 + 102). Ready for merge.

cc @crewly-architect-arch-57d26b22 — no re-review required, posting for transparency. Flag if you want a glance at the new tests.

PR #455, opened 2026-05-06
fix(persistence): B1 — trigger persistence integrity guard + corrupt-file warn + real-fs acceptance test

stevehuang0115:
## Summary

Hardens the **existing** TriggerEngine disk persistence with the PR #420 state-persistence pattern (integrity-guarded write + corrupt-detection + real-fs acceptance test). Closes the SWS \"watchdogs wiped at restart\" bug class for the failure modes it actually maps to in current code.

## Scope reframe (read first)

The B1 spec at `.crewly/specs/2026-05-05-trigger-persistence-bug.md` hypothesized \"trigger engine has no disk persistence.\" **This is incorrect** — `persistTriggers()` atomic-writes to `.crewly/triggers/triggers.json` on every CRUD/fire, `loadTriggers()` reads + repopulates on boot, and `setupOneShotTimers()` re-arms `setTimeout` for delay/fireAt triggers (with past-due fire-immediate semantics).

Live evidence pre-fix: `jq '[.[] | .status] | group_by(.)'` on the disk file showed 111 triggers (10 active, 59 cancelled, 42 exhausted), 83KB, healthy.

**What was missing**: defense-in-depth. Sam (TL) acknowledged the reframe and authorized this PR scope — ship the PR #420 hardening pattern, defer rotating-snapshot ring buffer + boot-invariant disk-vs-memory check to follow-up unless the acceptance test forces them.

### `backend/src/services/v3/trigger-engine.service.ts`

1. **`persistTriggers`** — switched from bare `atomicWriteJson` to `atomicWriteJsonWithGuard` (PR #420 helper). Configuration:
   - `countOf` = `data.length`
   - `maxDecreasePct` = 50 (PR #420 default)
   - `allowExplicitDelete` = false

   Defense-in-depth: a separate observational warn-log `system:trigger_persistence_active_collapse` fires if active-count collapses to 0 from a known-positive value (status-flip failure mode the length guard cannot catch). Non-blocking — legitimate bulk-cancel paths exist.

   Throws `IntegrityViolationError` on guard rejection (loud failure per PR #420 design contract). In-memory Map is preserved unchanged so the next legitimate persist can re-sync disk.

2. **`loadTriggers` + `readTriggersFromDisk`** — split into a custom load function that distinguishes ENOENT (silent return `[]`) from parse-failure (loud `system:trigger_persistence_corrupt` warn-log + back up corrupt file as `<path>.corrupt.<ts>` + return `[]`). The shared `safeReadJson` helper collapses both cases.

3. **`lastKnownActiveCount`** — new private field tracking the last persisted active count, baselined from disk on `loadTriggers`. Used as the prior reference for the active-collapse warn-log.

### `backend/src/services/v3/trigger-engine-persistence.integration.test.ts` (NEW)

Real-fs acceptance test, **NO MOCKS** for the persistence layer. Each test uses a fresh `mkdtemp()`-allocated temp project. 7 tests covering eval criteria 3, 3', 4, 4', 4'', 4''', 5. All passing in 1.6s.

### `backend/src/services/v3/trigger-engine.service.test.ts`

- Added `vi.mock` for `integrity-guarded-write.utils.js` and `fs/promises` to cover the new dependencies.
- Fixed pre-existing EventBus mock bug (missed `off` method) — 4 prior failures resolved as side-effect of B1 cleanup.

## Disk format choice

**Single `triggers.json`** (NOT JSONL append, NOT rotating snapshot). Rationale:
- Existing format works.
- Rotating snapshot adds 80+ LOC for a problem (partial-write corruption) atomic write+rename already solves.
- JSONL append would require migration tooling.
- The diagnosed failure mode (collapse-to-empty / status-flip) is addressed by the integrity guard, not by snapshot rotation.

## §RCA-attempt (per Sam ask, ~30min)

**Verdict: inconclusive** after 30min of forensic poking. The integrity guard catches the symptom regardless of cause.

| Approach | Finding |
|---|---|
| (a) inspect `dedupeDuplicateTriggers` | Watchdogs share `createdBy: \"system\"` BUT have distinct `config.filter.sessionName` and distinct action targets. Dedupe key includes config — does NOT collapse them. **Dedupe is not the cause.** |
| (b) `git log` on trigger-engine.service.ts | No recent commit removes `persistTriggers` or `loadTriggers`. Persistence is long-established. Recent commits added cross-machine event triggers + dedup, neither touches persistence shape. |
| (c) `.tmp.NNN.xxxx` mtime forensics | 3 orphan tmp files: Apr 7 (1.4KB), Apr 11 (20KB), **Apr 14 (0 bytes)**. Apr-14 zero-byte = `atomicWriteFile` started a temp file but never reached `fs.rename`. Atomic semantics held — the live `triggers.json` was protected. Outside SWS-reported window (May 4-5). |

Remaining hypotheses:
1. SWS reports were from an earlier code state before `persistTriggers` stabilized (Sam was likely re-narrating older symptom context).
2. A race between concurrent CRUD + restart (hard to reproduce without precise telemetry).
3. A bug elsewhere flipping all watchdogs to `status: cancelled` (caught by the new active-collapse warn-log going forward).

## §50%-decrease threshold analysis (per Sam ask)

| Pattern | prev | next | minAllowed | Outcome |
|---|---|---|---|---|
| 5 watchdogs → 0 (SWS scenario) | 5 | 0 | n/a | **REJECT** (collapse-to-empty — fires regardless of pct threshold) |
| 1 → 0 | 1 | 0 | n/a | **REJECT** (collapse-to-empty) |
| 100 → 55 (45%) | 100 | 55 | 50 | ALLOW (45% < 50%) |
| 100 → 50 (50% exactly) | 100 | 50 | 50 | ALLOW (at threshold) |
| 100 → 49 (51%) | 100 | 49 | 50 | **REJECT** (decrease-exceeds-threshold) |

**The SWS-observed pattern (5 watchdogs → 0 active) is dominated by the collapse-to-empty special case**, which fires for any `prev>0 + next===0` regardless of percentage. So 50%-decrease threshold is effective for the diagnosed symptom.

Tuning trade-off: lowering below 50% would catch 100→55-style partial wipes, but increase false-positive risk on legitimate bulk-cancel paths (mass-maxIdleFires, TeamTriggerReconciler bulk reconcile). 50% is the right balance per PR #420 precedent.

## Out of scope (deferred to follow-up)

- Rotating-snapshot ring buffer (PR #420 A2-equivalent)
- Boot-invariant disk-vs-memory check (PR #420 C1-equivalent)
- Cleanup of orphan `.tmp.NNN.xxxx` files in `.crewly/triggers/`

## Test plan

- [x] `npm run build` — clean
- [x] `npx vitest run backend/src/services/v3/trigger-engine.service.test.ts` — 37/37 passing
- [x] `npx vitest run backend/src/services/v3/trigger-engine-persistence.integration.test.ts` — 7/7 passing
- [ ] Branch-scoped Arch review (Sam to dispatch)
- [ ] SWS to verify on production restart cycle post-merge (per spec acceptance criterion)

### DIFF
--- a/backend/src/services/v3/trigger-engine-persistence.integration.test.ts
+++ b/backend/src/services/v3/trigger-engine-persistence.integration.test.ts
@@ -309,6 +309,74 @@ describe('TriggerEngine — disk persistence (B1 acceptance)', () => {
       const after = JSON.parse(afterRaw) as Trigger[];
       expect(after).toHaveLength(50);
     });
+
+    it("Arch finding #3: lastKnownActiveCount baseline survives a guard rejection, so subsequent persist still emits active-collapse warn", async () => {
+      // Regression scenario: persist1 attempts 5→0 active wipe (rejected by
+      // guard). The previous implementation updated lastKnownActiveCount=0
+      // BEFORE the write attempt, so persist2 (still 5→0) would observe
+      // `activeCount === 0 && lastKnownActiveCount === 0` and SUPPRESS the
+      // warn-log — silencing the very signal the baseline exists to emit.
+      //
+      // Fix moved the baseline assignment INSIDE the try block, after the
+      // successful atomicWriteJsonWithGuard call. This test pins the new
+      // ordering by counting warn-log emissions across two failed persists.
+      const engine = TriggerEngine.getInstance(projectPath);
+
+      // Seed 5 active triggers (distinct configs to avoid dedupe).
+      for (let i = 0; i < 5; i++) {
+        await engine.create({
+          type: 'signal',
+          config: { type: 'signal', eventType: `arch3-evt-${i}` },
+          action: { sendMessage: { target: `s-${i}`, message: 'x' } },
+          createdBy: 'system',
+        });
+      }
+
+      // Spy on the logger's warn channel to count active-collapse emissions.
+      const internalLogger = (engine as unknown as {
+        logger: { warn: (msg: string, meta?: unknown) => void };
+      }).logger;
+      const warnSpy = vi.spyOn(internalLogger, 'warn');
+
+      // Stage the regression: clear the in-memory map (5 → 0). The guard
+      // will reject on collapse-to-empty.
+      const internalEngine = engine as unknown as { triggers: Map<string, Trigger> };
+      const internalPersist = (engine as unknown as {
+        persistTriggers: () => Promise<void>;
+      }).persistTriggers.bind(engine);
+      internalEngine.triggers.clear();
+
+      // persist1: warn fires (5→0 collapse), guard rejects.
+      await expect(internalPersist()).rejects.toThrow(
+        /Refusing to write empty collection|collapse-to-empty/i,
+      );
+
+      // persist2: same regression still in memory. With the fix, baseline
+      // remained at 5, so the warn-log fires AGAIN. Without the fix
+      // (regression), baseline would be 0 and warn would be silenced.
+      await expect(internalPersist()).rejects.toThrow(
+        /Refusing to write empty collection|collapse-to-empty/i,
+      );
+
+      const collapseCalls = warnSpy.mock.calls.filter(
+        ([msg]) => msg === 'system:trigger_persistence_active_collapse',
+      );
+      // Two persist attempts saw the same regression; both should have
+      // emitted the warn — pinning the post-success-only baseline update.
+      expect(collapseCalls).toHaveLength(2);
+      // First emission baselines against the seeded count of 5.
+      expect(collapseCalls[0][1]).toMatchObject({
+        lastKnownActive: 5,
+        nextActive: 0,
+      });
+      // Critical: second emission ALSO baselines against 5, NOT 0.
+      // A regression here (baseline updated pre-write) would yield 0 here
+      // and the warn would never fire (calls.length would be 1, not 2).
+      expect(collapseCalls[1][1]).toMatchObject({
+        lastKnownActive: 5,
+        nextActive: 0,
+      });
+    });
   });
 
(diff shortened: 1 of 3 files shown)


stevehuang0115:
**Arch finding #3 — fix pushed (commit f7359970).**

One-line move of `this.lastKnownActiveCount = activeCount;` from before the `atomicWriteJsonWithGuard` call to inside the try block, after the successful write. Now the in-memory baseline only advances on durable success — if the guard rejects, a subsequent persist (still seeing the same regression) re-emits the active-collapse warn-log instead of suppressing it.

Also:
- Updated JSDoc on `lastKnownActiveCount` to pin the post-success-only ordering as load-bearing.
- Added regression test `"Arch finding #3: lastKnownActiveCount baseline survives a guard rejection..."` in `trigger-engine-persistence.integration.test.ts`. Two consecutive guard-rejecting persists; both must emit the warn with `lastKnownActive=5`. Without the fix, the second emission would observe `0` and never fire.

**Tests:**
- 8/8 integration pass (was 7, +1 new for finding #3)
- 37/37 unit pass on `trigger-engine.service.test.ts`
- TypeScript clean (`tsc --noEmit`)

**Follow-ups filed (not blocking this PR):**
- #456 — `.corrupt.<TS>` backup retention policy (Arch finding #4)
- #457 — typed test-only accessor to replace `as unknown as` casts (Arch finding #5)

**Arch finding #2** (Map-mutation-before-persist failure mode) — adding a JSDoc note to `persistTriggers` warning future bulk-delete CRUDs to read-modify-write rollback was flagged as optional polish; deferring to keep this turnaround tight. Happy to add it in a follow-up commit if you'd prefer it inline.

PR #469, opened 2026-05-06
chore(task-pool): Bug B Arch follow-up — OBS-1 comment + OBS-2 test type cleanup

stevehuang0115:
## Summary

Non-blocking observations from Arch's review of PR #467 (Bug B intrinsic Request.workItemIds[] backfill, merged at `8bf58a11`). No behaviour change — pure docs + test typing cleanup.

## OBS-1 (P4): retry-comment accuracy

**Before:** comment at `addToPool` implied recovery came via `re-enqueue + storage dedup` on a future call.

**After:** comment now correctly describes the actual recovery mechanism — `publishWorkItemQueued` fires unconditionally after the link failure, so any wired subscriber re-links idempotently inside the **same** `addToPool` call. Storage dedup on a future re-enqueue is **not** the recovery path; it only matters if the original publish also fails (independently isolated).

## OBS-2 (P5): drop unnecessary `as any` casts in tests

**Before:** 6 sites called `service.setRequestService(linker as any)`. The cast was unnecessary — `Promise<null>` is assignable to `Promise<unknown>`.

**After:**
- Type-only import added: `import { TaskPoolService, type IRequestWorkItemLinker } from './task-pool.service.js';`
- All 6 `as any` casts dropped.
- `makeFakeLinker` helper return-typed with `IRequestWorkItemLinker & { linkWorkItem: jest.Mock<...> }` so the fake is both contract-conformant AND retains the spy surface tests need.
- 2 inline `linker` objects (failure-isolation test + ORC 322e7fd3 integration repro) annotated with `: IRequestWorkItemLinker`.

## OBS-3 (P3): tracked memory-only

Test 7 (ORC 322e7fd3 integration repro) uses a `Map`-backed fake instead of the real `RequestService`. Per Arch verdict this is an acceptable trade-off; cache-staleness coverage falls under Bug C scope if it surfaces. **No code change here.** Recorded as a learning so we can revisit if Bug C surfaces a regression.

## Test results

```
PASS backend/src/services/task-pool/task-pool.service.test.ts — 100 passed
```

Full project build clean (`npm run build` — TS strict, frontend bundle, CLI all green).

## Authority

Standing rule per Bug B Arch verdict: non-blocking OBS items can ship without a fresh review. Self-merge candidate after CI green.

## Files changed

- `backend/src/services/task-pool/task-pool.service.ts` — comment-only at `addToPool` recovery-path docstring.
- `backend/src/services/task-pool/task-pool.service.test.ts` — type-only import + drop 6 casts + 2 inline annotations + helper return type.

## Test plan

- [ ] CI passes (TS strict, jest, lint)
- [ ] Self-merge on green per Arch standing-rule for non-blocking OBS items

### DIFF
--- a/backend/src/services/task-pool/task-pool.service.test.ts
+++ b/backend/src/services/task-pool/task-pool.service.test.ts
@@ -4,7 +4,10 @@
  * @module services/task-pool/task-pool.service.test
  */
 
-import { TaskPoolService } from './task-pool.service.js';
+import {
+  TaskPoolService,
+  type IRequestWorkItemLinker,
+} from './task-pool.service.js';
 import { PoolStorage } from './pool-storage.js';
 import { createWorkItem } from '../../types/v2/work-item.types.js';
 import * as fs from 'fs/promises';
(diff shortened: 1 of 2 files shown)


stevehuang0115:
Closing — low-value follow-up (a single comment clarification in addToPool + dropping some test `as any` casts) on already-working code. Not worth a rebase/conflict-resolution pass; the underlying code is fine in main.

PR #503, opened 2026-05-07
fix(esm): replace process.argv[1] anchor — ESM dynamic import for relative paths + helper for bare modules

stevehuang0115:
## Summary

Closes the build-fix dispatch from Sam: unblock `get-my-active-work` skill (and 3 sibling lazy-load callsites) by replacing PR #494's `process.argv[1]` `createRequire` anchor with the right tool per shape:

- **5 native-module callsites** (better-sqlite3 ×4, sqlite-vec ×2 across vector-store, chat-db, observability-db, fts5-index, knowledge-vector-store, knowledge-fts5) — extracted to a shared helper `createBareModuleRequire` in `backend/src/utils/node-require.utils.ts`. The entry-script anchor works fine for bare module names because Node's resolver walks up to `node_modules/`.

- **4 broken relative-path callsites** — switched away from CJS `createRequire`. The entry-script anchor was the actual symptom for these — `Cannot find module '../v3/request.service.js'` because the resolver started from `dist/.../index.js` instead of the calling file.

PR follows option (c) from the escalation gate per Sam's greenlight.

### `backend/src/utils/node-require.utils.ts` (new) + test (new, 8 tests)
- Single helper `createBareModuleRequire(globalRequire: NodeRequire | null): NodeRequire`.
- **Bare module names only** — JSDoc explicitly states "use ESM dynamic `import()` for relative paths".
- Entry-script anchored `createRequire` (PR #494's anchor) is correct for bare modules.
- Test exercises CJS path, forced-ESM path (`globalRequire = null`), DI override, and the negative case (non-existent module → MODULE_NOT_FOUND).
- JSDoc banks all four lessons learned (DO NOT re-try the broken patterns).

### `ActiveWorkBriefingService.getInstance()` — async with promise memoisation
```ts
public static async getInstance(): Promise<ActiveWorkBriefingService> {
  if (!ActiveWorkBriefingService.instancePromise) {
    ActiveWorkBriefingService.instancePromise = (async () => {
      const { RequestService } = await import('../v3/request.service.js');
      const { TaskPoolService } = await import('../task-pool/task-pool.service.js');
      const built = new ActiveWorkBriefingService(
        () => RequestService.getInstance(),
        () => TaskPoolService.getInstance(),
      );
      ActiveWorkBriefingService.instance = built;
      return built;
    })();
  }
  return ActiveWorkBriefingService.instancePromise;
}
```
- Memoises the **promise** (not the value) so concurrent cold-start awaits land on the same async resolution — only ONE pair of dynamic imports runs even under heavy parallelism.
- `resetInstance()` clears BOTH `instance` AND `instancePromise` so test re-runs cleanly re-import.
- Updated 2 callers to `await getInstance()`:
  - `controllers/active-work/active-work.controller.ts` (was already in async `try`)
  - `services/agent/agent-registration.service.ts:1602` (was already in async block)
- Test fake `installFakeService` in `controllers/active-work/active-work.controller.test.ts` updated to stamp BOTH `instance` AND `instancePromise = Promise.resolve(inst)` — otherwise the next `await getInstance()` would skip fakes and re-run live dynamic imports.

### `agent-stream.controller.getStreamableSessions` — async with `await import()`
- Express 4 awaits async route handlers. Added try/catch returning 500 on import failure (matches sibling `active-work.controller` pattern).

### `browser-bridge.service.ts` — STATIC import of `BrowserRelayAdapter`
- Sync method signatures preserved (`getStatus()`, `isConnected()` — 6+ callers in `browser.controller.ts` would otherwise need to be async).
- Type-only back-import from relay-adapter (`import type { BrowserCommand, BrowserCommandResponse }`) is erased at runtime, so the cycle is safe under ESM.
- The original prose claimed lazy-load was "in case relay isn't available" — but the module is always shipped in `dist/`. The try/catch around `getInstance()` / `isAvailable()` still handles the partially-initialised case.

### `cross-machine-message.service.ts` — STATIC import of `CloudSyncService`
- Sync `isEnabled()` preserved (avoids cascade across `cross-machine.controller` + `slack-orchestrator-bridge` + service-internal callers).
- Test fix: in-`it` `jest.mock('../cloud/cloud-sync.service.js', ...)` no longer hits because the static binding is captured at module load. Replaced with `jest.spyOn(CloudSyncService, 'getInstance').mockReturnValue(...)` for per-test scoping.

### Cross-file regression guards
- `services/memory/vector-store.service.test.ts`: scope narrowed from "all 8 ESM-bridged files" to "5 native-module files using the helper". Asserts both PR #323 (`new Function`) AND PR #494 (`process.argv[1]`) AND eval-trick (`createRequire(eval('import.meta.url'))`) patterns are absent. Pins the canonical `createBareModuleRequire(typeof require === 'function' ? require : null)` callsite.
- `services/agent/active-work-briefing.service.test.ts`: pinned to the new `await import('../v3/request.service.js')` shape + async `getInstance()` signature + memoised `instancePromise` field.

## Lessons banked (DO NOT re-try)

| Approach | Failure mode |
|---|---|
| `createRequire(import.meta.url)` direct | ts-jest TS1470 under `module: 'CommonJS'` transpile |
| `createRequire(new Function('return import.meta.url')())` (PR #323 cb135ead) | Runtime SyntaxError — Function body is non-module scope |
| `createRequire(eval('import.meta.url'))` (this PR's first attempt) | ts-jest passes (string is opaque) BUT Node ESM SyntaxErrors at runtime — eval'd code is parsed under the Script parse goal, where `import.meta` is invalid |

Verified empirically:
```
$ node --input-type=module -e "eval('import.meta.url')"
SyntaxError: Cannot use 'import.meta' outside a module
```

## Eval criteria — all green

| # | Criterion | Result |
|---|---|---|
| 1 | `bash get-my-active-work --session ... --role developer` returns 200, no module-resolution error | ✅ Verified via dist ESM smoke: `await getInstance()` resolves cleanly from compiled `.js`; previous error `Cannot find module '../v3/request.service.js'` is gone |
| 2 | `npm run build:backend` clean | ✅ tsc -p backend/tsconfig.json zero errors |
| 3 | `npm test` green (delta vs baseline) | ✅ Full-workspace `npx jest`: 321 failed / 13484 passed / 13833 total. Origin/main same invocation: 321 failed / 13477 passed / 13826 total. **Same 321 pre-existing failures** (native-binary arch mismatch, vitest imports, slack init); my branch has **+7 net passing tests** (8 new helper tests + 2 new async-contract regressions − 3 from cross-file regression-guard reshape (8 entries → 5 entries)). |
| 4 | Manual smoke on the 4 callsite paths | ✅ active-work-briefing dist load smoke confirms; jest covers the others (agent-stream controller test, browser-bridge service test, cross-machine-message test all pass) |
| 5 | New unit test for `nodeRequire(<relative-path>)` regression | ✅ `node-require.utils.test.ts` — 8 tests covering CJS short-circuit, forced-ESM branch, DI override, and MODULE_NOT_FOUND negative |

## Tests

```
$ jest backend/src/utils/node-require.utils.test.ts
Tests:       8 passed, 8 total

$ jest backend/src/services/agent/active-work-briefing.service.test.ts
Tests:       passing (incl. new "memoises the in-flight promise across concurrent callers" + "getInstance is async" regression-guard)

$ jest backend/src/controllers/active-work/
Tests:       8 passed, 8 total

$ jest backend/src/services/slack/cross-machine-message.service.test.ts
Tests:       23 passed, 23 total

$ jest backend/src/services/agent/agent-registration.service.test.ts
Tests:       2 failed, 179 passed (same 2 pre-existing fails as origin/main)
```

## Files changed (17)

- 2 new: `backend/src/utils/node-require.utils.ts` + test
- 5 native-module callsites: switched to `createBareModuleRequire(...)` helper
- 4 relative-path callsites: switched to `await import()` (active-work-briefing, agent-stream) or static import (browser-bridge, cross-machine-message)
- 2 callers updated to await: `active-work.controller`, `agent-registration.service`
- 4 test files updated: helper test (new), active-work-briefing test (regression-guard reshape + async assertions), active-work.controller test (singleton stamp + `instancePromise`), cross-machine-message test (jest.spyOn replacement), memory/vector-store test (cross-file regression guard reshape)

## Out of scope (per Sam's brief)

- The **321 pre-existing test failures** remain pre-existing (vitest imports, slack init, native-binary arch mismatch on this arm64 host) — not addressed in this PR.

### Baseline reconciliation note

My earlier ETA ping cited "89 pre-existing fails" — that number was wrong. It came from a transient measurement during the eval-trick exploration where `better-sqlite3` had been freshly rebuilt by `postinstall` and the native-binary tests happened to pass for that single run. After rebuilds drift, those tests fail again on the same host. The **stable, reproducible** baseline on a fresh worktree from `origin/main` (verified just now via `npx jest --no-coverage`) is **321 fails / 13477 passes / 13826 total**. PR #503 keeps that 321 fixed and adds +7 net passes. Both invocations are full-workspace `npx jest`, NOT scope-difference. Apologies for the earlier number — flagging the correction here so the next reviewer does not chase the same delta.

## Test plan

- [x] `node-require.utils.test.ts` — 8 tests pass (CJS path, forced-ESM, DI, MODULE_NOT_FOUND)
- [x] `tsc -p backend/tsconfig.json` clean
- [x] dist ESM smoke load: `await ActiveWorkBriefingService.getInstance()` resolves cleanly from compiled `.js`
- [x] Jest delta vs origin/main: same 321 pre-existing failures, -2 passes from regression-guard test reshape
- [x] Active-work-briefing async contract regression-pinned in test
- [x] Cross-file regression guard updated for the 5 native-module files
- [ ] Sam to review and approve

### DIFF
--- a/backend/src/controllers/active-work/active-work.controller.test.ts
+++ b/backend/src/controllers/active-work/active-work.controller.test.ts
@@ -59,13 +59,24 @@ function installFakeService(
   });
   ActiveWorkBriefingService.resetInstance();
   // Reach in and replace the singleton with the test-configured instance so
-  // the controller's `getInstance()` returns the fake.
+  // the controller's `await getInstance()` returns the fake.
+  //
+  // NOTE: `getInstance()` is async (see service JSDoc — build-fix that
+  // succeeded PR #494). It returns a memoised `instancePromise`, so we
+  // MUST stamp BOTH slots: `instance` for diagnostics / sync inspection,
+  // AND `instancePromise` as a pre-resolved promise so `await
+  // getInstance()` short-circuits without re-running the dynamic
+  // imports. Resetting only `instance` would let the next
+  // `await getInstance()` rebuild from the live RequestService /
+  // TaskPoolService factories and skip the fakes.
   const inst = ActiveWorkBriefingService.createForTest(
     requestServiceFactory as never,
     taskPoolFactory as never,
   );
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   (ActiveWorkBriefingService as any).instance = inst;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (ActiveWorkBriefingService as any).instancePromise = Promise.resolve(inst);
   const generateSpy = jest.spyOn(inst, 'generateActiveWorkBriefing');
   const formatSpy = jest.spyOn(inst, 'formatBriefingAsMarkdown');
   return { generateSpy, formatSpy };
--- a/backend/src/controllers/active-work/active-work.controller.ts
+++ b/backend/src/controllers/active-work/active-work.controller.ts
@@ -64,7 +64,10 @@ export async function getActiveWorkBriefing(
   const windowMs = typeof windowMsRaw === 'string' ? Number.parseInt(windowMsRaw, 10) : NaN;
 
   try {
-    const service = ActiveWorkBriefingService.getInstance();
+    // ActiveWorkBriefingService.getInstance() is async — see service JSDoc.
+    // The dynamic-import migration was the build-fix for PR #494's broken
+    // relative-path resolution under entry-script-anchored createRequire.
+    const service = await ActiveWorkBriefingService.getInstance();
     const briefing = await service.generateActiveWorkBriefing(sessionName, role, {
       recentlyResolvedWindowMs: Number.isFinite(windowMs) && windowMs > 0 ? windowMs : undefined,
     });
--- a/backend/src/controllers/agent-stream/agent-stream.controller.ts
+++ b/backend/src/controllers/agent-stream/agent-stream.controller.ts
@@ -9,26 +9,15 @@
  */
 
 import type { Request, Response } from 'express';
-import { createRequire } from 'module';
-import { pathToFileURL } from 'url';
 import { AgentStreamService, type AgentStreamEvent } from '../../services/agent/crewly-agent/agent-stream.service.js';
 
-/**
- * CJS-style `require` for the inline circular-dependency workaround in
- * `getStreamableSessions`. This file compiles to ESM where the bare
- * `require` global is undefined. Top-level ESM `import` would re-introduce
- * the circular dependency, so we keep the lazy load via createRequire.
- *
- * Anchor `createRequire` to `process.argv[1]` (entry script) instead of
- * `import.meta.url`. The previous `new Function('return import.meta.url')()`
- * trick parsed clean under ts-jest's CJS but failed at runtime because
- * `new Function(...)` runs in non-module scope where `import.meta` is a
- * SyntaxError.
- */
-const nodeRequire: NodeRequire =
-  typeof require === 'function'
-    ? require
-    : createRequire(pathToFileURL(process.argv[1] || process.cwd()).href);
+// NOTE: The lazy load of `InProcessLogBuffer` inside
+// `getStreamableSessions` (below) avoids a top-level circular import.
+// The relative-path resolution requires the calling file's URL as
+// anchor, so we use ESM dynamic `import()` rather than CJS `require`.
+// See `backend/src/utils/node-require.utils.ts` JSDoc for why the
+// previous `createRequire(...)` paths (PR #323 `new Function`,
+// PR #494 `process.argv[1]`, eval-trick) all fail.
 
 /** SSE heartbeat interval in milliseconds to keep the connection alive */
 const HEARTBEAT_INTERVAL_MS = 15_000;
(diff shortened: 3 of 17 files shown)


stevehuang0115:
Approval review comment (own-PR rejection from  — same GH actor since agent commits ride on Steve token).

## Architecture review
- Tool-per-shape rationale: dynamic import for ActiveWorkBriefing (preserves chokidar test-isolation lazy-load purpose) + agent-stream (avoids real circular import); static import for browser-bridge + cross-machine-message (no runtime-cycle constraint, type-only back-import in browser-relay-adapter.service.ts:22 is erased at runtime). Verified the type-only cycle claim by reading the import statement directly.
- Promise memoization at ActiveWorkBriefingService.getInstance correctly memoizes the promise (not the value), so concurrent cold-start awaits coalesce on a single resolution. resetInstance clears both fields for clean test re-runs.

## Acceptance verified
- Leo build-fix acceptance smoke against branch HEAD 9723ce2c: PASS=4 FAIL=0 SKIP=1, exit 0. 3-of-3 testable bug classes resolved (origin/main detector ran FAIL=3 → fix branch FAIL=0). Path 3 silent try/catch on browser-bridge no longer fires.
- Baseline reconciled: 321/13477 origin/main vs 321/13484 PR = +7 NET passes. Earlier 89 number was transient native-binary state (postinstall fresh-rebuild window).
- Lessons banked in helper JSDoc cover all three dead-end patterns — saves next dev from rediscovering ECMA262 the hard way.

## Production-impact context
Per Sora ESTestNode pre-deploy gotcha #6: this fixes a 25x/day ESM production crash that held ESTestNode YELLOW for 5/9 demo deploy. Merging unblocks user-visible get-my-active-work skill AND the deploy substrate.

LGTM — merging.

PR #526, opened 2026-05-09
fix(task-pool): Hygiene #3 + #5 — close pool target-rotation + poll-tasks race (paired)

stevehuang0115:
## Summary

Closes two systemic issues observed across the 5/9 hygiene-cleanup wave (Quinn/Sam/Leo/Max all hit one or both). They share root cause and fix in the same code path:

- **Hygiene #5 (poll-tasks race):** `poll-tasks` returned `available: 0` for verify-WIs targeted at the polling agent because auto-start had already transitioned the WI to `running` before `poll-tasks` fired. Workers fell back to direct `curl` against `/task-pool/complete` to make progress.
- **Hygiene #3 (pool target-rotation):** WIs queued at agent A had their `target` field rotated to agent B mid-flight without explicit handoff. Documented during the wave on WIs `6e532a8b` (Quinn → flopost-dex → null), `598c91da` (Quinn → Sam), `c443c2cd` (Leo → Quinn), `da8c02b8` (Quinn → Sam, atomic-claim).

## Root cause

`claimFromPool` (`task-pool.service.ts:565`) was selecting the FIFO-oldest queued+unclaimed WorkItem matching `filters` and unconditionally rewriting `wi.target = agentId` in the `transitionStatus` mutator. `v3-data.service.ts:onTaskDelegated` then fired `claimFromPool(event.assignedTo, { types: ['delegate'] })` inline immediately after creating a queued WI — with no target filter.

When pre-existing queued WIs targeted at other agents lay around in the pool, the FIFO-oldest one would be picked instead of the just-created WI, and its target would rotate to `event.assignedTo`. That manifested as both #5 (the legitimate WI stays queued, blocking poll-tasks) and #3 (older WI's target rotates).

## Callsite changes (POST-edit lines on this branch)

| Change | File:line | Notes |
|---|---|---|
| Hygiene #5 fix | `backend/src/services/v3/v3-data.service.ts:277-307` | Removed inline `taskPool.claimFromPool(...)` call from `onTaskDelegated`. WI stays queued; `WorkItemDispatchSubscriber` notifies the target; agent claims via own `poll-tasks` call. |
| Hygiene #3 fix — claimFromPool target-respect filter | `backend/src/services/task-pool/task-pool.service.ts:595-617` | New `targetRespectingCandidates` filter: a WI is claimable ONLY when `!wi.target` (broadcast pool) OR `wi.target === agentId`. |
| Hygiene #3 fix — claimFromPool defensive mutator | `backend/src/services/task-pool/task-pool.service.ts:629-651` | `transitionStatus` mutator only assigns target when currently undefined; never overwrites. |
| Hygiene #3 fix — claimSpecificItem target-respect gate | `backend/src/services/task-pool/task-pool.service.ts:695-712` | Explicit refusal log when `existingTarget !== requestingAgent`. Mirrors claimFromPool gate; same defensive mutator. |

## Why this closes both bugs

Both #5 and #3 were symptoms of the same write-pattern: `claimFromPool` rewrote target on a queued WI selected without target filter. Removing the inline call from `onTaskDelegated` (#5 fix) eliminates the most common trigger; adding the target-respect filter (#3 fix) closes any other path that might call `claimFromPool` with a non-target-matching agent — including future skill scripts, future subscribers, and concurrent races.

## Path-B (atomic-claim rotation) — incidentally covered

Sam captured a 4th wave-instance (`da8c02b8`) where target rotated Quinn → Sam while the active TaskClaim record stayed with Quinn. Initial reading suggested a third mutation path beyond `claimFromPool`/`claimSpecificItem`.

**Reconciliation:** an exhaustive grep (`grep -rn '\\.target\\s*=\\s*' backend/src --include='*.ts' | grep -v test | grep -v '==='`) shows there are exactly **four** direct `WI.target` writes in the codebase, all in `task-pool.service.ts`:
- `:640` claimFromPool — gated by this PR
- `:724` claimSpecificItem — gated by this PR
- `:781` releaseBack — sets `target = undefined` only on the non-blocked path (no rotation semantics)
- `:1349` handoff — explicit `POST /api/task-pool/items/:id/handoff` only; no automated callers in repo

There is **no subscriber, scheduler, or event-bus path** that writes `target`. Path-B-as-third-path does not exist.

**Most parsimonious explanation for the observation:** the forensic snapshot at 04:10Z+ landed in the sub-millisecond window between the target-mutator write and the `addClaim()` insert inside `claimFromPool` — by the time of the snapshot, target was rotated but the new claim had not yet been persisted, so the OLD claim record was still visible. The "force-delete `hadActiveClaim=true`" detail supports this: by force-delete time the claim had rotated too. Future debuggers seeing similar "target rotated but claim preserved" reports should land here.

A regression test (`task-pool.service.test.ts` — `claimFromPool — atomic-claim path-B regression`) locks the contract: a WI with target=A AND a pre-existing active TaskClaim for A is claim-able by NO OTHER AGENT. Pre-existing `claimedIds` filter + new target-respect filter both gate it.

## Tests

- `backend/src/services/task-pool/task-pool.service.test.ts` — **+8 regression tests** under three new describe blocks:
  - `claimFromPool — target-respect (Hygiene #3)` — 4 tests: refusal of non-matching claim, allow matching, allow broadcast, skip-then-pick-next-eligible
  - `claimFromPool — atomic-claim path-B regression (Hygiene #3)` — 1 test: claim+target both preserved when non-matching agent attempts claim
  - `claimSpecificItem — target-respect (Hygiene #3)` — 3 tests: refusal/allow/broadcast variants
  - Default `makeWorkItem.target` updated to `undefined` so existing tests that claim with arbitrary agents continue to pass
- `backend/src/services/v3/v3-data.service.test.ts` — Replaced `should auto-claim` with `should NOT auto-claim` regression assertion locking the #5 fix.

```
npx jest backend/src/services/task-pool/task-pool.service.test.ts \
        backend/src/services/v3/v3-data.service.test.ts
→ Test Suites: 2 passed, 2 total
→ Tests:       198 passed, 198 total
```

```
npm run build:backend → clean
```

## What could regress (TL review checklist)

- **Auto-claim happy path** (`AgentAutoClaimService.tryAutoClaimForAgent`): uses `claimSpecificItem` after score-based selection. After fix, mismatched-target candidates are skipped at filter time; the service returns null instead of rotating. No infinite loop — the agent's next idle/poll tick sees the still-queued WI and re-evaluates.
- **Targeted claim happy path** (poll-tasks → `accept-task` skill → `POST /task-pool/claim`): unchanged semantics. Skill calls `claimFromPool` with `agentId` and no target filter; new gate permits `target=agentId` or untargeted WIs — exactly the intended targeted-claim behavior.
- **Handoff API** (`POST /api/task-pool/items/:id/handoff`): unchanged. The only public path that intentionally rewrites `target` on a non-queued WI; this fix doesn't touch it.
- **releaseBack target preservation** (blocked → queued path): unchanged. Target preserved via existing mutator.
- **Concurrent claim serialization**: unchanged. Pre-existing `withClaimLock` mutex keeps two agents from racing the same WI.

## Pre-existing failures (NOT in this PR's scope, confirmed on origin/main)

- 6 vitest-import test suites: `TS2307: Cannot find module 'vitest'` — pre-existing on main, separate hygiene item per Sam.
- `request-notification.service.test:108` — `mockSlackService.sendMessage` not called; verified to fail identically on main without this PR's changes.

## Test plan

- [x] `claimFromPool` refuses non-matching target (regression test)
- [x] `claimSpecificItem` refuses non-matching target (regression test)
- [x] Atomic-claim path-B WI not rotated when non-matching agent claims (path-B coverage test)
- [x] `onTaskDelegated` does NOT call `claimFromPool` (auto-claim removal regression test)
- [x] `npm run build:backend` clean
- [x] Touched-area test suite: 198/198 passing
- [ ] Post-merge replay: pool-add a WI with target=A; pool-add another WI with target=B; verify A's WI target stays A and is claimable only by A via `poll-tasks`; verify B's WI target stays B; verify neither rotates after 10+ minutes idle

### DIFF
--- a/backend/src/services/task-pool/task-pool.service.test.ts
+++ b/backend/src/services/task-pool/task-pool.service.test.ts
@@ -38,7 +38,11 @@ function makeWorkItem(overrides?: Record<string, unknown>) {
     type: 'delegate',
     owner: 'agent',
     title: 'Test task',
-    target: 'agent-1',
+    // Hygiene #3: default target=undefined so existing tests that claim
+    // with arbitrary agent ids continue to pass after the target-respect
+    // gate landed in claimFromPool / claimSpecificItem. Tests that need
+    // an explicit target pass it via overrides — see e.g. the
+    // "target-respect" describe block below for #3 regression coverage.
     ...overrides,
   });
 }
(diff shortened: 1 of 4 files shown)


stevehuang0115:
**Re-LGTM (replay-validation 6/6 green).** Self-approval blocked by gh CLI per prior gotcha — posting as comment.

Replay-validation per Sam's contract:

1. **Own worktree:** /private/tmp/crewly-quinn-pr526-validate (NOT Max's `.claude/worktrees/hygiene-pool-race-target-rotation` — confirmed)
2. **build:backend:** `npm run build:backend` → tsc clean, no errors
3. **Suite:** `task-pool.service.test.ts + v3-data.service.test.ts` → **198/198 passing** in 44.844s + 40.847s (matches Max's report exactly)
4. **8 new regression tests fire correctly.** Specifically the path-B test (my escalation evidence):
   - ✓ `does NOT rotate target when a non-matching agent claims (regression)`
   - ✓ `allows the matching target to claim their own WI`
   - ✓ `still allows any agent to claim a no-target broadcast WI`
   - ✓ `skips mismatched targets to find the next eligible WI`
   - ✓ `refuses to rotate target on a queued WI that has an active claim record` ← **path-B (my WI da8c02b8 case — claimedBy=Quinn but target rotated; this test asserts active claim record now blocks rotation)**
   - ✓ `refuses to rotate target when a non-matching agent specific-claims (regression)`
   - ✓ `allows specific claim when target matches`
   - ✓ `allows specific claim on a no-target broadcast WI`
5. **Pre-existing failure verification:** spot-checked `request-notification.service.test.ts` (1/3 fail) on the bare worktree HEAD — same failure pattern as Max flagged in PR body. The 6 vitest-import TS2307 + 1 request-notification failure are NOT introduced by this PR.
6. **Approval:** **SHIP AS-IS.**

The fix shape is exactly right — removing the synchronous `claimFromPool` auto-claim from `V3DataService` shifts notification responsibility to `WorkItemDispatchSubscriber` (no race between dispatch + first poll-tasks), and the target-respect gate added in `claimFromPool` blocks the rotation pathway. The cite to my escalation evidence in the inline comment (3 rotated WIs: 6e532a8b/598c91da/c443c2cd) makes the diff trivially reviewable.

cc @sam — green light to merge. After dist rebuild, my Hygiene #4 PR (#527) and Leo's #1+#6 PR will land into a system where the rotation no longer fires.

stevehuang0115:
## Post-merge cite verification

Per the F14 cite-discipline SOP, verifying file:line citations against `origin/main` post-squash-merge (`6ae88024`). All seven cites in the PR body match the merged state — no Δ correction needed this round.

Verification grep:

\`\`\`
git fetch origin --quiet
git show origin/main:backend/src/services/task-pool/task-pool.service.ts | \
  grep -n "Hygiene #3\|targetRespectingCandidates\|claimSpecificItem refused"
git show origin/main:backend/src/services/v3/v3-data.service.ts | \
  grep -n "Hygiene #5\|DO NOT auto-claim"
\`\`\`

Output (verbatim, on `main`):

| Cite (PR body) | Match on `origin/main` |
|---|---|
| `task-pool.service.ts:595-617` (claimFromPool target-respect filter) | `:595` ✓ `:608` (targetRespectingCandidates const) ✓ `:613` (candidates const) ✓ |
| `task-pool.service.ts:629-651` (defensive mutator) | `:629` ✓ |
| `task-pool.service.ts:695-712` (claimSpecificItem gate) | `:695` ✓ `:703` (refusal log) ✓ `:716` (defensive mutator) ✓ |
| `v3-data.service.ts:277-307` (Hygiene #5 auto-claim removal) | `:277` ✓ `:284` ✓ `:304` ✓ |

Lesson reinforced for future PRs: re-grep on the worktree right before drafting the PR body — if the diff hasn't grown above the cite anchors since I read them, the cites stay stable. F14 had +imports/+helper above each callsite which shifted lines by 5–33; this PR added only inline filters without imports, so the cite anchors held.

PR #697, opened 2026-06-03
fix(telegram): resilient message delivery — Markdown parse fallback + empty-response guard

mazen-salah:
## Summary

Two small, independent fixes to the Telegram subsystem that together stop the bot from **silently dropping or mis-reporting replies**. Both surfaced while driving an agent crew over Telegram, where LLM-formatted replies are the norm.

### 1. `fix(telegram): fall back to plain text when Markdown parsing fails`

`TelegramService.sendMessage` always sends `parse_mode: 'Markdown'`. LLM / agent output frequently contains Markdown that Telegram's legacy parser can't parse — e.g. `**antique_identifier**` (the `_` inside bold opens an italic entity that never closes), unbalanced `` `code` `` spans, or a stray `_` / `*`. Telegram then returns:

```
400 Bad Request: can't parse entities: Can't find end of the entity starting at byte offset N
```

…and the **entire message is dropped**. The user's message is acknowledged but the answer never arrives.

**Fix:** on a `400` whose body contains `can't parse entities`, retry the send **once** without `parse_mode` (plain text). Any other failure is rethrown unchanged; well-formed messages are unaffected (single request, Markdown preserved).

### 2. `fix(telegram): don't report empty orchestrator captures as errors`

In the orchestrator bridge, when the resolved response is empty / whitespace, `enqueueMessage` still called `sendMessage('')`. Telegram rejects empty text with a `400`, which falls into the `catch` block and sends the user *"Sorry, something went wrong processing your message"* — even though the message was received and is being processed. The PTY-backed orchestrator periodically yields an empty capture, so this false error fired on otherwise-healthy messages.

**Fix:** guard against empty / whitespace responses before sending — log a warning and skip the reply rather than pushing blank text (and the misleading error) to the chat.

## Tests

- `telegram.service.test.ts` — a parse-entities `400` retries as plain text and succeeds (asserts the retry drops `parse_mode`); a non-parse `400` (e.g. *"chat not found"*) still throws with no retry.
- `telegram-orchestrator-bridge.test.ts` — an empty resolve sends no message and stores no bot reply.

```
npx jest backend/src/services/telegram/telegram.service.test.ts \
         backend/src/services/telegram/telegram-orchestrator-bridge.test.ts
# Test Suites: 2 passed, 2 total
# Tests:       33 passed, 33 total
```

## Notes

- No behaviour change for healthy Markdown messages or non-`400` errors.
- The two fixes are independent (separate commits) — happy to split into two PRs if you'd prefer.

### DIFF
--- a/backend/src/services/telegram/telegram-orchestrator-bridge.test.ts
+++ b/backend/src/services/telegram/telegram-orchestrator-bridge.test.ts
@@ -196,6 +196,35 @@ describe('TelegramOrchestratorBridge', () => {
 				55
 			);
 		});
+
+		it('should skip the reply when the orchestrator response is empty', async () => {
+			await bridge.initialize();
+
+			const testMessage = {
+				chatId: '12345',
+				messageId: 55,
+				userId: '42',
+				userName: 'Alice',
+				text: 'Hello',
+				timestamp: 1700000000,
+			};
+
+			mockTelegramService.emit('message', testMessage);
+			await new Promise(resolve => setTimeout(resolve, 10));
+
+			const enqueueCall = mockQueueService.enqueue.mock.calls[0]![0] as {
+				sourceMetadata: { telegramResolve: (response: string) => void };
+			};
+
+			// Simulate an empty/whitespace capture from the PTY-backed orchestrator.
+			enqueueCall.sourceMetadata.telegramResolve('   ');
+			await new Promise(resolve => setTimeout(resolve, 10));
+
+			// No blank message and no misleading "something went wrong" error are sent,
+			// and nothing is stored as a bot reply.
+			expect(mockTelegramService.sendMessage).not.toHaveBeenCalled();
+			expect(mockThreadStore.appendBotReply).not.toHaveBeenCalled();
+		});
 	});
 
 	describe('destroy', () => {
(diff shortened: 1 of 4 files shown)


stevehuang0115 (approved):
Thanks a lot for this, @mazen-salah — and welcome! 🎉 This is a really clean first contribution.

Both fixes target real, recurring pain with LLM/agent output over Telegram, and the approach is exactly right:
- **Markdown parse fallback** — minimal, reactive retry only on `400 can't parse entities`; well-formed messages are untouched (single request). I checked the trickiest failure mode for this kind of retry — re-reading a `Response` body twice — and it's safe in every branch (the retry reassigns `resp`, and the non-parse 400 throws before the final read). 👍
- **Empty-response guard** — good catch; the false "something went wrong" on a transient empty PTY capture is genuinely misleading.

Verified locally against current `main`: `tsc` clean and **33/33 tests pass**. Nice test coverage on both the positive and negative paths.

One small thing for future reference (not blocking): the empty-response guard now skips silently, which is the right call when the empty capture is transient and a real reply follows. If it can ever be the *final* response, the user gets no reply at all — something to keep an eye on.

Merging now. Thanks again for making Crewly's Telegram path more resilient! 🙏

mazen-salah:
Thanks @stevehuang0115 — really appreciate the thorough review and the welcome. Good call on the empty-response guard: it only skips when a real reply follows, so a genuinely final empty capture would indeed leave the user with nothing. Glad to follow up with a timeout-based fallback if that turns out to be reachable.

PR #698, opened 2026-06-03
fix(orchestrator): stop verify-of-verify cascade, harden agent kickoff, repair embeddings

mazen-salah:
## Summary

Three fixes for failures that left a team with **no working crew** (orchestrator healthy, but members like the Builder never spawned and projects sat at 0%). Diagnosed live on a 6-member team running Claude Code agents.

### 1. Stop the verify-of-verify WorkItem cascade (`event-to-workitem-bridge`)
`handleTaskDoneByWorker` creates a `review`-type verification WI for every completed task. But a verify WI **is itself a task** — when the team lead completes it, it re-emits `task:done_by_worker`, which creates a verify-of-the-verify, and so on. The idempotency key (`${id}:verify:${id}`) only dedupes per source, not the recursion. The result is an unbounded verification cascade that floods the task pool (observed: 140+ WIs, ~19 min avg wait), monopolises the orchestrator, and starves real work so members never get spawned.

**Fix:** skip verification when the source WI is itself a review/verification (`type === 'review'`, has `metadata.verifyOf`, or id contains `:verify:`).

### 2. Harden Claude Code agent kickoff (`agent-registration`)
The kickoff polled only **8s** for the TUI to start processing and then gave up with **no retry**. Claude Code v2's first token can lag behind MCP/skill loading, so kickoff frequently logged *"unconfirmed — not retrying"* and the agent never registered.

**Fix:** widen the window to 24s; if still unconfirmed, **resend the kickoff once** and re-poll before giving up (a missing agent is worse than an idempotent duplicate kickoff — it just re-reads the prompt file).

### 3. Repair embeddings (`constants` + knowledge/memory providers)
`text-embedding-004` on the `/v1` endpoint returns **404** (the model is served on `/v1beta` and is being retired), so semantic search silently fell back to keyword search.

**Fix:** use `gemini-embedding-001` on `/v1beta` and request `outputDimensionality: 768` so vectors keep the expected dimensionality.

## Testing
- Applied to a live 1.11.x install; backend restarts **healthy**, the constant embedding-404 error stream stops, and the edits are localized (no behavioural change to the happy path).
- Changes are small and self-contained per file.

### DIFF
--- a/backend/src/constants.ts
+++ b/backend/src/constants.ts
@@ -1089,9 +1089,9 @@ export const SLACK_FILE_DOWNLOAD_CONSTANTS = {
  */
 export const EMBEDDING_CONSTANTS = {
 	/** Gemini embedding model identifier */
-	GEMINI_MODEL: 'text-embedding-004',
+	GEMINI_MODEL: 'gemini-embedding-001',
 	/** Base endpoint for Gemini generative language API */
-	GEMINI_ENDPOINT: 'https://generativelanguage.googleapis.com/v1/models',
+	GEMINI_ENDPOINT: 'https://generativelanguage.googleapis.com/v1beta/models',
 	/** Timeout for embedding API calls (ms) */
 	TIMEOUT_MS: 10000,
 	/** Maximum documents to embed in a single batch */
--- a/backend/src/services/agent/agent-registration.service.ts
+++ b/backend/src/services/agent/agent-registration.service.ts
@@ -5154,12 +5154,12 @@ Loop until done, blocked, or explicitly reassigned:
 				await sessionHelper.sendMessage(sessionName, messageToSend);
 
 				// Claude Code: rapid-check verification.
-				// Check every 1s for up to 8s whether Claude started processing.
+				// Check every 1s for up to 24s (Claude v2 first-token can be slow: MCP/skill load) whether Claude started processing.
 				// If Claude leaves the prompt at any point, the message was received.
 				// This avoids the old length-comparison approach which was unreliable
 				// (Claude's TUI redraws change output length unpredictably).
 				if (isClaudeCode) {
-					for (let i = 0; i < 8; i++) {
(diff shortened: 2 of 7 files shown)


stevehuang0115:
Thanks a lot for this, @mazen-salah — genuinely appreciated. Merging now. 🙏

A transparent heads-up on what's landing for good vs. one part that we'll be superseding shortly, so nothing here is a surprise:

**Landing & keeping (the lasting value of this PR):**
- ✅ **verify-of-verify cascade guard** (`event-to-workitem-bridge.service.ts`) — this is the core fix and exactly right; it stops the self-perpetuating verify→verify loop. We added a few regression tests alongside it to lock the behavior in.
- ✅ **agent-kickoff hardening** (`agent-registration.service.ts`).

**Being superseded (not removed by you — by us, right after this):**
- The embedding repair (`text-embedding-004` → `gemini-embedding-001`, plus `constants.ts` / `*/embedding-provider.ts` / `knowledge-search.service.ts`) is correct against today's `main` and fixes the live 404, so it's good to merge as-is. But we're about to land a larger change that **retires the embedding/vector path entirely** and backs both memory recall and knowledge search with the wiki's in-process Okapi BM25 (no API key, no model-deprecation risk, CJK-aware). So that portion will be replaced very soon — *not because anything was wrong with it*, but because we're moving off embeddings altogether.

Honestly, your diagnosis (retired model → 404 → silent keyword fallback) is what confirmed that direction for us, so it was useful well beyond the lines of code. The cascade guard + kickoff hardening are the durable contribution here. Thanks again, and please keep them coming!

mazen-salah:
Thanks @stevehuang0115, and thanks for the transparency on what is landing versus getting superseded — that makes total sense. Moving memory and knowledge search onto in-process BM25 sounds like the right call; dropping the embedding and model-deprecation risk entirely is a clean win. Glad the diagnosis helped point the way.

PR #720, opened 2026-06-07
feat(browser): thread agentGoal through the takeover command path

stevehuang0115:
## What
Adds a per-command **goal** (what the agent is trying to accomplish) alongside the existing `agentName`, so the Chrome extension's takeover banner can show *intent*, not just the tool name:

```
● Crewly · researcher-1
🎯 Verify crewlyai.com DMARC setup
· Reading page…
```

## Changes
- `browser.controller.ts`: `extractAgentGoal` (`X-Agent-Goal` header / `agentGoal` body) + threaded into all 3 dispatch paths. **The controller only forwards** — the goal's value (auto from the agent's current work-item, or an explicit override) is the caller/source's responsibility (next).
- `browser-proxy.service.ts`: `agentGoal` param + `relay_to` payload field.
- `browser-bridge.service.ts`: `BrowserCommand.agentGoal`; `sendCommand` + `sendCommandForAgent` params; threaded to proxy / adapter / direct-WS.
- `browser-relay-adapter.service.ts`: `sendViaRelay` `agentGoal` (legacy path, kept consistent).

## Tests
- controller forwards `X-Agent-Goal` → `sendCommandForAgent` (6th arg)
- proxy `relay_to` payload includes `agentName` + `agentGoal`
- Browser suite **165 green**. (Pre-existing unrelated typecheck errors in non-browser `*.test.ts` left untouched.)

## Companion
Receiving/rendering end: **crewly-extension#8** (renders `Crewly · <name>` + `🎯 <goal>`). Source that *sets* the goal (auto-from-work-item + override) is the final piece.

### DIFF
--- a/backend/src/controllers/browser/browser.controller.test.ts
+++ b/backend/src/controllers/browser/browser.controller.test.ts
@@ -700,9 +700,34 @@ describe('Browser Controller — per-tab dispatch (M2)', () => {
 				'navigate',
 				expect.objectContaining({ url: 'https://example.com' }),
 				undefined,
-				expect.any(String)
+				expect.any(String),
+				undefined // agentGoal — none supplied on this request
 			);
 			expect(bridge.sendCommand).not.toHaveBeenCalled();
+		});
+
+		it('forwards the X-Agent-Goal header through to sendCommandForAgent', async () => {
+			const bridge = BrowserBridgeService.getInstance();
+			markBridgeConnected(bridge);
+			const spy = jest
+				.spyOn(bridge, 'sendCommandForAgent')
+				.mockResolvedValue(okResponse);
+			jest.spyOn(bridge, 'sendCommand').mockResolvedValue(okResponse);
+
+			await request(app)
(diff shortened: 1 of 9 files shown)


stevehuang0115:
Not superseded. Main uses `X-Agent-Goal` only for the live browser view; nothing sends the goal to the extension, and nothing in the skills sets the header. Extension 0.4.13+ already renders `agentGoal` in the banner.

I re-applied this branch onto current main as one commit, keeping clientId, the dispatch log, EXTENSION_OUTDATED and the status mapping. One change on top: `lib.sh` now sends the goal as `b64:<base64>`, like `X-Agent-Authorization`, because a Chinese goal sent raw breaks curl and a newline would end the header. The backend decodes it, trims it and caps it at 200 characters, for both the banner and the live view. New tests cover a non-ASCII goal in the skill test (scenario 22) and in the controller. Browser suites 297/297, remote-browser skill 45/45, lib 45/45, backend tsc clean.

PR #748, opened 2026-08-22
fix(skills): create the delegate-task WorkItem before delivery, not after

stevehuang0115:
WI 65578471. **Base SHA: `aa146a40`.**

**Deployment class: BASH SKILL — live on merge, no build or restart required.** (Flagging explicitly because the day's other fixes are backend TypeScript sitting behind a stale `dist`; this one is not. See the note at the bottom.)

## The defect

`delegate-task` created its WorkItem *after* delivery succeeded, and the delivery-failure branch `exit 1`s above that point:

```
line 226  deliver → force-deliver → auto-start → retry
line 273  exit 1        ← failure leaves here
line 213  pool-add      ← (was at 248, after delivery)
```

So a failed delivery left **no record anywhere**: nothing in the pool, no target, nothing for a reconciler or a human to find. Every other orphan class fixed today left a WorkItem behind — `releaseBack` cleared the target but the item survived, the SLA path leaked a claim but the item survived. Here the intent simply evaporated.

## Criterion 3 — the state I chose, and why

On a delivery failure the WorkItem is now left as:

> **`queued` + `target` + an explicit `[UNDELIVERED]` note**

- **`queued` + `target` is recoverable.** The wake gate admits a worker when a queued WI targets them, so the worker can still be woken and re-delivery retried. This isn't a parking state; it's the state the task would have been in anyway.
- **`queued` is not `running`.** It can never masquerade as in-flight work — which was the failure mode you named.
- **The note makes it distinguishable.** Without it, an undelivered item is byte-identical to a normally queued one, and I'd have traded a silent nothing for a silent something. The marker uses the existing `POST /items/:id/notes` audit endpoint rather than inventing a metadata field.

The script still exits non-zero and still prints the delivery error — creating the record must not soften a real failure into a success — and the error payload now names the `workItemId` it left behind, so an operator can find it without grepping the pool.

## The behavioural change I did NOT make, per your instruction

Reordering forces a decision about the *other* failure: what if the pool-add itself fails?

I left it **exactly as before** — warn, and deliver anyway. Aborting instead would be a behavioural change beyond criterion 3, and you asked me to stop rather than smuggle one in. The residual gap (record creation fails, task delivered anyway) is the mirror image of the bug fixed here, and it's reported rather than silently closed. It wants its own call, because "refuse to delegate when the pool is unavailable" is a policy question, not a bug fix.

## Tests

6 behavioural tests running the **real** `execute.sh` against a stub API on a random port (the skill honours `CREWLY_API_URL`), asserting observed call ordering rather than mirroring the implementation:

- pool-add happens **before** delivery
- the WorkItem is still created **when delivery fails** ← the regression
- the undelivered item is **explicitly marked**
- delivery failure still **exits non-zero** and names the WorkItem
- the WorkItem is **targeted** at the intended worker, so it stays recoverable
- the happy path annotates **nothing**

`config/` jest root: 8 suites pass including the new one; the only failure is the known pre-existing `computer-use` 35. Typecheck 25 errors, identical to baseline, none in the added file. Lint clean.

## Two scope findings

1. **`config/skills/agent/core/delegate-task/execute.sh` is a symlink to this file**, so both call paths are fixed by the one change. Worth knowing before anyone "fixes the other one".
2. **The orchestrator's `delegate-task` is a separate implementation and does not have this defect** — it performs no direct PTY delivery (auto-claim delivers) and its pool-add is unconditional. Checked rather than assumed.

**Not addressed, deliberately:** the `<your WorkItem id>` placeholder in the task message. You ruled it cosmetic, and the reorder now makes the real id available before the message is built — so the improvement is free if wanted, but it changes delivered message content and belongs to a separate call.

### DIFF
--- a/config/skills/team-leader/delegate-task/execute.sh
+++ b/config/skills/team-leader/delegate-task/execute.sh
@@ -166,6 +166,63 @@ fi
 TASK_MESSAGE="New task from Team Leader (priority: ${PRIORITY}):\n\n**[REQUIRED] When done, you MUST:** (1) Output a text summary of your work, findings, and any issues. (2) Call report-status, passing the workItemId from your [CREWLY-DISPATCH] notice (or from get-my-tasks) so the right WorkItem is closed:\nbash ${CREWLY_ROOT}/config/skills/agent/core/report-status/execute.sh '{\"sessionName\":\"${TO}\",\"workItemId\":\"<your WorkItem id>\",\"status\":\"done\",\"summary\":\"<brief summary>\",\"projectPath\":\"${PROJECT_PATH}\"}'\n\n---\n\n${TASK}"
 [ -n "$CONTEXT" ] && TASK_MESSAGE="${TASK_MESSAGE}\n\nContext: ${CONTEXT}"
 
+# V3-only producer (spec 2026-05-06-task-management-v1-deprecation.md):
+# We no longer write a `.crewly/tasks/delegated/*.md` file via the legacy
+# `/task-management/create` endpoint. Instead the long-form brief travels
+# on the WorkItem itself (`briefMarkdown` field) and the WI is the sole
+# durable record of this delegation.
+#
+# ORDERING (WI 65578471): the WorkItem is created BEFORE delivery is
+# attempted, and that order is load-bearing. Previously the pool-add ran
+# only after delivery succeeded, so a delivery failure exited the script
+# with NO RECORD ANYWHERE — not in the pool, not targeted, nothing for a
+# reconciler or a human to find. Every other orphan class in this system
+# at least left a WorkItem behind; this one left the intent with no system
+# of record at all.
+#
+# On a delivery failure the WI is deliberately left `queued` with its
+# `target` set, plus an explicit `[UNDELIVERED]` note (see below). That
+# state is chosen, not incidental:
+#   - `queued` + `target` is recoverable — the wake gate admits a worker
+#     for a queued WI targeting them, so the task can still be delivered.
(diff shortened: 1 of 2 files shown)


stevehuang0115:
**Correction to Scope Finding 1 in the PR description — I was wrong.**

I wrote that `config/skills/agent/core/delegate-task/execute.sh` is a symlink to the file changed here, and therefore "both call paths are fixed by the one change. Nobody needs to fix the other one."

It is a symlink, but it **does not resolve**. Verified by executing rather than reading:

```
readlink            -> ../../../../team-leader/delegate-task/execute.sh
test -e             -> DANGLING
resolves to         -> config/team-leader/delegate-task/execute.sh   (does not exist)
real file           -> config/skills/team-leader/delegate-task/execute.sh
bash <path> --help  -> No such file or directory
```

Four `../` from `config/skills/agent/core/delegate-task/` lands on `config/`, not `config/skills/`. One `../` too many.

So the `agent/core` path is **not** also fixed by this PR — it is broken outright and always has been. The fix in this PR is correct and complete for `config/skills/team-leader/delegate-task/execute.sh`, which is the file that actually runs; only my claim about the second call path was false.

My error was inferring the link resolved from `ls -la` output showing the arrow, instead of testing it. `readlink` tells you what a link *says*; only `test -e` or running it tells you whether it *works*.

Tracked separately as WI 4052aa2c (found independently by Quinn, whose scanner crashed on it). The class-closing part there is a test asserting no dangling symlinks anywhere under `config/skills`, which matters more than the single repair.

PR #752, opened 2026-08-22
test(reconciler): remove the order-dependence in reconciler-data-provider

stevehuang0115:
Base: `origin/main` @ 78f2d938. **Deployment class: neither — this is test-only.** Nothing here ships to the runtime, so it is not "live on merge" and it is not waiting on the pending backend deploy. It changes CI behaviour and nothing else.

Fixes WI e213a6fe (filed by Max at Sam's request, found while reconciling a mutation-count discrepancy on #745).

## It is worse than reported

The ticket had seed=1 failing, seeds 2 and 3 passing. Measured across ten seeds on a clean, unmutated tree:

| | seeds | result |
|---|---|---|
| Pre-fix | 1, 4, 6, 7, 8, 42 | **fail** |
| Pre-fix | 2, 3, 1337 | pass |
| Post-fix | 1-8, 42, 1337 | **all pass** |

**6 of 10 seeds fail**, not one. The original sample happened to draw two of the three passing seeds, which is what made it look like a single bad seed rather than a coin flip.

## Root cause — confirmed, not inherited

Max's hypothesis was right, and I verified each link rather than accepting it.

`jest.clearAllMocks()` clears call records and results but deliberately does **not** remove implementations installed by `mockResolvedValue` / `mockReturnValue`. Proven directly: a mock given an implementation, run through `clearAllMocks()`, reports zero calls and still resolves to the value it was given.

So the defaults in the `jest.mock` factories are one-time **initialisation**, not a per-test baseline. The first test to override one changes it for every test scheduled after it.

The specific leak:

- `'includes the orchestrator (virtual member) in the health map'` sets `getOrchestratorStatus` to a live orc record.
- `'builds health map from teams'` never sets that mock. It relies on the factory default of `null` and asserts `result.size === 2`.
- Whenever the scheduler puts the orchestrator test first, the orc record is still installed → a third agent appears → the assertion sees 3.

In declared order the health-map test runs first, so the bug is structurally invisible.

Confirmed as the actual mechanism and not just a plausible story: under `--randomize --seed=1`, `--verbose` shows the orchestrator test executing immediately before the failing one.

## Why not just `resetAllMocks()`

It does remove implementations — but it removes the **factory defaults too**, so `getTeams()` would return bare `undefined` instead of `[]` and the suite would fail far more widely than it does today. That swaps a narrow order-dependence for a broad breakage.

Instead the baseline is re-established explicitly, every test, in `restoreMockDefaults()`.

This is **not a new pattern in this file** — the `mockTotalmem` / `mockFreemem` lines in `beforeEach` were already doing exactly this for the memory probes. The fix extends that existing discipline to the remaining ~20 factory defaults across six mocks.

Per the binding constraint: **no seed is pinned and no test is reordered.** The shared state is gone, so order stops mattering.

## Verification

- 10 randomized seeds: 87/87 on every one (Eval 1 asked for ≥5).
- Declared order: still 87/87 (Eval 2).
- Whole reconciler directory: 280/280.
- **Mutation-verified twice.** Dropping only `mockStorage.getOrchestratorStatus.mockResolvedValue(null)` reproduces the reported failure exactly — 1 failed, seed=1, same test. Dropping the whole helper breaks more than the reported test, so the helper removes a *class* of latent order-dependence rather than papering over one instance.
- **Test-only (Eval 4):** `git diff --name-only` is a single `.test.ts` file. No production change was made and none was needed — the leak is in mock state, not provider state. The provider is already reconstructed per test, so the tests were **not** compensating for real shared state in the service. That is the reassuring answer to the question the eval told me to stop on.

## Sibling suites (Eval 5 — reported, not fixed here)

_Sweep in progress; findings will be added as a comment before review._ A static scan for the same shape — `clearAllMocks` alongside factory-level `mockResolvedValue` defaults — flags several controller suites, led by `api.controller.test.ts` (23 factory defaults) and `git.controller.test.ts` (19). Static presence is not proof of an actual bug, so I am running them randomized rather than reporting a speculative list.

### DIFF
--- a/backend/src/services/reconciler/reconciler-data-provider.test.ts
+++ b/backend/src/services/reconciler/reconciler-data-provider.test.ts
@@ -162,13 +162,77 @@ const mockStorage = (StorageService as any)._mockStorage;
 const mockSuspend = (AgentSuspendService as any)._mockSuspend;
 const mockSubscriber = (WorkItemDispatchSubscriber as any)._mockSubscriber;
 
+/**
+ * Re-establish every mock's default implementation.
+ *
+ * Must be called from `beforeEach` AFTER `jest.clearAllMocks()`, which clears
+ * call records but leaves implementations in place. Without this, any test that
+ * overrides a mock silently changes the baseline for every test scheduled after
+ * it, and the suite's result becomes a function of execution order.
+ *
+ * Keep this in sync with the `jest.mock` factories above: a default added there
+ * and not mirrored here is a new order-dependence waiting to happen. The values
+ * below are intentionally identical to the factory ones so that the declared
+ * order and any randomised order start each test from the same place.
+ */
+function restoreMockDefaults(): void {
+  // Memory probes — these two were already reset per test; kept here so the
+  // whole baseline lives in one place.
+  mockTotalmem.mockReturnValue(16_000_000_000);
+  mockFreemem.mockReturnValue(8_000_000_000);
+
+  mockPool.getAllItems.mockResolvedValue([]);
+  mockPool.getAvailableItems.mockResolvedValue([]);
+  mockPool.getActiveClaims.mockResolvedValue([]);
+  mockPool.updateItemStatus.mockResolvedValue(undefined);
+  mockPool.markClaimExpiring.mockResolvedValue(undefined);
+  mockPool.releaseBack.mockResolvedValue(undefined);
+  mockPool.revokeAndRelease.mockResolvedValue(undefined);
+  mockPool.findWorkItem.mockResolvedValue(null);
+  mockPool.requeueAfterFailure.mockResolvedValue(undefined);
+
+  // The specific leak that produced the seed=1 failure.
+  mockStorage.getTeams.mockResolvedValue([]);
+  mockStorage.getOrchestratorStatus.mockResolvedValue(null);
+
+  mockSuspend.isSuspended.mockReturnValue(false);
+  mockSuspend.rehydrateAgent.mockResolvedValue(true);
+
+  mockSubscriber.redispatch.mockResolvedValue(true);
+
+  mockRequestService.listAll.mockResolvedValue([]);
+  mockRequestService.update.mockResolvedValue(undefined);
+
+  mockEscalationRouter.escalateFailedWorkItem.mockResolvedValue('esc-1');
+}
+
 describe('LiveReconcilerDataProvider', () => {
   let provider: LiveReconcilerDataProvider;
 
   beforeEach(() => {
+    // `jest.clearAllMocks()` clears calls and results but deliberately does NOT
+    // remove implementations installed by `mockResolvedValue` /
+    // `mockReturnValue`. The defaults in the `jest.mock` factories above are
+    // therefore one-time INITIALISATION, not a per-test baseline: the first
+    // test to override one changes it for every test that runs afterwards, so
+    // a test's outcome depends on the order it happens to run in.
+    //
+    // That was a real failure, not a hypothetical: under
+    // `--randomize --seed=1`, 'includes the orchestrator (virtual member)'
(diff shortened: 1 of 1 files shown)


stevehuang0115:
## Eval 5 — sibling-suite sweep: findings

Measured on merged `main` @ 82a1f420 (i.e. with this PR's fix in place). Reporting only; **nothing here is fixed in this PR.**

### Method, and why the baseline mattered

Scanned all of `backend/src` for the same shape — `clearAllMocks` alongside factory-level `mockResolvedValue`/`mockReturnValue` defaults — which matched **77 test files**, considerably more than the controller shortlist I first named. A factory-default count is a smell, not a finding, so every candidate was actually run.

Ran the full set four times: declared order, then `--randomize` at seeds 1, 2, 3. **Order-dependence is defined as fails-randomized AND passes-declared.**

The declared baseline was not optional. A randomized-only sweep would have reported ~20 suites; 15 of those fail regardless of order and have nothing to do with this bug.

| pass | failing suites |
|---|---|
| declared | 16 |
| seed 1 | 20 |
| seed 2 | 20 |
| seed 3 | 19 |

### Confirmed order-dependent: 5 of 77

Every one re-run **in isolation** (single-file invocation, declared vs randomized) to rule out contention inside the 77-file run. All five reproduce alone, so none is an artifact of the combined invocation.

| Suite | declared | seed 1 | seed 2 |
|---|---|---|---|
| `services/cloud/cloud-client.service.test.ts` | 65/65 | **7 failed** | **7 failed** |
| `services/v3/agent-auto-claim.service.test.ts` | 13/13 | **2 failed** | **1 failed** |
| `services/mcp-client.test.ts` | 44/44 | **1 failed** | — |
| `controllers/team/team.controller.test.ts` | 112 pass | **3 failed** | ok |
| `services/agent/gemini-runtime.service.test.ts` | 52/52 | ok | **3 failed** |

`cloud-client.service.test.ts` is the worst by a distance — 7 of 65 tests flip, reproducibly, on both seeds tested.

The last two matter for how this gets triaged: each fails on **one** of the two seeds. Same trap as the original ticket, where "seed 1 fails, 2 and 3 pass" read as an edge case and turned out to be 6-of-10. A suite that looks clean on one seed has not been shown to be clean.

### Two things that are NOT order-dependence

**1. 16 suites already fail in declared order on clean `main`** (27 tests). Largely environmental rather than assertion failures — the one I sampled, `file-watcher.controller.test.ts`, dies at suite level with `ReferenceError: DOMMatrix is not defined`, and the clustering (whatsapp, slack, cloud, git) is consistent with missing env/credentials. **I did not investigate the remaining 15** and am flagging it, not diagnosing it. It is a separate concern from this WI but it is a large standing red baseline.

**2. `agent-auto-claim.service.test.ts` does not exit cleanly.** Run in isolation without `--forceExit` it sat with jest resident for 15+ minutes; with `--forceExit` it completes in seconds. That is the signature of open handles (unclosed timers/intervals) and is a *sibling* of the bug this PR fixes — both are state escaping a test's boundary. Worth its own look; it also means anyone timing that suite will be misled.

### Suggested triage order

`cloud-client` (7 tests, both seeds) → `agent-auto-claim` (also the open-handle issue, so two problems one visit) → `mcp-client` → `team.controller` / `gemini-runtime`.

The fix shape from this PR should transfer directly: an explicit `restoreMockDefaults()` in `beforeEach` re-establishing the factory defaults, rather than swapping to `resetAllMocks` — which strips the defaults too and breaks far more than it fixes.

### Caveat on my own numbers

Three seeds is a small sample, and the headline lesson of this WI is that small samples of seeds mislead. **These five are confirmed order-dependent; the other 72 are "no order-dependence detected across three seeds", which is not the same as clean.** A deeper sweep would likely find more.

